PR #2, opened 2023-03-21
Added enableAntiAffinity in instance spec

kaikash:
Add enableAntiAffinity flag for instanceSpec, so it's possible to enable anti affinity for any node (not only for master nodes). Also, might be useful to add whole corev1.Affinity in InstanceSpec.


savnadya:
Hi, I've merged your commit into our internal repository. It will appear in Github in about an hour. Thank you!

PR #20, opened 2023-07-11
fix: store location 'medium' property should be named as 'medium_name' in yson

dpoluyanov:


maxim-babenko:
We must also make sure that these media are actually being created at master during cluster init.

PR #23, opened 2023-08-01
Add Support for Different Media Types to ytsaurus Kubernetes Operator

nordikafiles:
This pull request introduces the ability to handle different media types in the ytsaurus Kubernetes operator. 

### DIFF
--- a/api/v1/ytsaurus_types.go
+++ b/api/v1/ytsaurus_types.go
@@ -260,11 +260,14 @@ type DeprecatedSpytSpec struct {
 	SpytVersion  string `json:"spytVersion,omitempty"`
 }
 
+type Medium struct {
+	Name string `yson:"name"`
+}
+
 // YtsaurusSpec defines the desired state of Ytsaurus
 type YtsaurusSpec struct {
-	CoreImage string `json:"coreImage,omitempty"`
-	UIImage   string `json:"uiImage,omitempty"`
-
+	CoreImage        string                        `json:"coreImage,omitempty"`
+	UIImage          string                        `json:"uiImage,omitempty"`
 	ImagePullSecrets []corev1.LocalObjectReference `json:"imagePullSecrets,omitempty"`
 	ConfigOverrides  *corev1.LocalObjectReference  `json:"configOverrides,omitempty"`
 	AdminCredentials *corev1.LocalObjectReference  `json:"adminCredentials,omitempty"`
--- a/pkg/components/init_job.go
+++ b/pkg/components/init_job.go
@@ -4,9 +4,10 @@ import (
 	"context"
 	"fmt"
 	"path"
+	"strings"
+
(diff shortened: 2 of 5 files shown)


pkg/components/init_job.go
```diff
@@ -30,6 +31,17 @@ func initJobWithNativeDriverPrologue() string {
 	return strings.Join(commands, "\n")
 }
 
+func createChangePrimaryMediumForTablesCommands(mediumName string, replicationFactor int, tables []string) string {
+	commands := []string{}
+
+	for _, table := range tables {
+		commands = append(commands, "/usr/bin/yt set "+table+"/@media/ssd_blobs  '{\"replication_factor\" = 3; \"data_parts_only\" = %false;}'")
```
psushin, line 38: I would have leaved default replication factor, that is specifically adjusted by operator for small testing clusters, having just one or two data nodes.

Event if you want to adjust replication factor manually, I would suggest to use ```@replication_factor``` attribute, instead of ```@media```

pkg/components/master.go
```diff
@@ -120,6 +133,17 @@ func (m *master) createInitScript() string {
 		fmt.Sprintf("/usr/bin/yt set //sys/@cluster_connection '%s'", string(clusterConnection)),
 		"/usr/bin/yt set //sys/controller_agents/config/operation_options/spec_template '{enable_partitioned_data_balancing=%false}' -r -f",
 		m.initAdminUser(),
+		m.initMedia(),
+	}
+
+	if m.cfgen.HasSsdBlobs() {
```
psushin, line 139: 1. since `@primary_medium` is an inheritable attribute, it'is better to set the medium on the directory, and not to hardcode all table paths.
2. I think it's better to add desired medium for operations_archive to ytsaurus spec explicitly

psushin (changes requested):

pkg/ytconfig/generator.go
```diff
@@ -480,3 +480,25 @@ func (g *Generator) GetDiscoveryConfig() ([]byte, error) {
 	}
 	return marshallYsonConfig(c)
 }
+
+func (g *Generator) GetExtraMedia() []v1.Medium {
```
savnadya, line 484: This method isn't connected with `ytconfig`, let's move it to `master.go`.

api/v1/ytsaurus_types.go
```diff
@@ -260,11 +260,14 @@ type DeprecatedSpytSpec struct {
 	SpytVersion  string `json:"spytVersion,omitempty"`
 }
 
+type Medium struct {
```
savnadya, line 263: This struct is also used only in `master.go`, let's move it too.

savnadya:
It was fixed [here](https://github.com/ytsaurus/yt-k8s-operator/commit/7094bdfc7266a74af80d983990ae008c89bc8893).

PR #28, opened 2023-09-10
Fix ENABLE_WEBHOOKS behavior

zlobober:

Before this commit ENABLE_WEBHOOKS=false does not stop disable CHYT webhooks. This commit achieves the desired behavior.


savnadya:
Shipped!

savnadya:
Shipped!

PR #30, opened 2023-09-10
YTNEBIUS-27: request privileged mode for exec node containers by default

zlobober:

This commit makes possible to perform privileged actions on behalf of exec-node, in particular - mount tmpfs in jobs, which currently results in user slot disabling.


savnadya:
Shipped!

PR #31, opened 2023-09-13
Make privileged field optional

zlobober:

Fix after https://github.com/ytsaurus/yt-k8s-operator/pull/30


savnadya:
Shipped!

PR #34, opened 2023-09-15
Support specifying Odin webservice

zlobober:

After this commit, a hacky way of hardcoded Metrika counters for public is removed. On the other hand, it becomes possible to specify UI custom config.


savnadya:
Shipped!

PR #35, opened 2023-09-24
Various improvements for working in host network mode.

zlobober:
- Introduce high-level hostNetwork option; for now it affects everything except masters.
- Resolve port clashes for nodes of different flavors.
- Temporarily specify skynet ports; better to remove them in the future.

### DIFF
--- a/pkg/consts/address.go
+++ b/pkg/consts/address.go
@@ -16,8 +16,19 @@ const (
 	ControllerAgentRPCPort        = 9014
 	ControllerAgentMonitoringPort = 10014
 
-	NodeRPCPort        = 9012
-	NodeMonitoringPort = 10012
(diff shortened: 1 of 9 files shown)


pkg/consts/address.go
```diff
@@ -16,8 +16,19 @@ const (
 	ControllerAgentRPCPort        = 9014
 	ControllerAgentMonitoringPort = 10014
 
-	NodeRPCPort        = 9012
-	NodeMonitoringPort = 10012
+	DatNodeRPCPort        = 9012
```
koct9i, line 19: NodeFlavor* без сокращений, тут тоже нет смысла

savnadya (changes requested):

PR #37, opened 2023-10-06
Introduce postprocess-config init container

zlobober:
After this commit config files for server components are not taken as is from config map, but rather treated as templates, which are additionally postprocessed by embedding environment variables like POD_NAME and POD_NAMESPACE.

### DIFF
--- a/pkg/components/volume.go
+++ b/pkg/components/volume.go
@@ -1,10 +1,12 @@
 package components
 
 import (
+	"fmt"
 	ytv1 "github.com/ytsaurus/yt-k8s-operator/api/v1"
 	"github.com/ytsaurus/yt-k8s-operator/pkg/consts"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"path"
 )
 
 func createVolumeClaims(specVolumeClaimTemplates []ytv1.EmbeddedPersistentVolumeClaim) []v1.PersistentVolumeClaim {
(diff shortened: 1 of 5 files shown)


pkg/components/volume.go
```diff
@@ -22,24 +24,33 @@ func createVolumeClaims(specVolumeClaimTemplates []ytv1.EmbeddedPersistentVolume
...
 func createVolumeMounts(specVolumeMounts []v1.VolumeMount) []v1.VolumeMount {
 	volumeMounts := make([]v1.VolumeMount, 0, len(specVolumeMounts)+1)
 	volumeMounts = append(volumeMounts, specVolumeMounts...)
+	volumeMounts = append(volumeMounts, createConfigTemplateVolumeMount())
 	volumeMounts = append(volumeMounts, createConfigVolumeMount())
 	return volumeMounts
 }
 
-func createConfigVolume(configMapName string, mode *int32) v1.Volume {
+func createConfigTemplateVolume(volumeName string, configMapName string, mode *int32) v1.Volume {
```
savnadya, line 51: It's used for `consts.ConfigTemplateVolumeName` and for `consts.ConfigVolumeName`. So I think `createConfigVolume` is better name for this function.

savnadya (changes requested):

pkg/components/volume.go
zlobober: Fixed

savnadya:
Shipped!

PR #52, opened 2023-11-12
Fix deadlock in case when UI deployment is manually deleted

zlobober:

When somebody tries to delete UI deployment while cluster is in UpdateStateWaitingForPodsRemoval, current definition of arePodsRemoved returns false, while logically pods are indeed removed if there is no deployment. It results in a deadlock which prevents UI from reappearing again.

Also, it looks like the same issue may happen with server components, so the similar change is also done for them.


savnadya:
Shipped!

PR #54, opened 2023-11-14
Support dualstack clusters in YQL agent

zlobober:
Without this change it is impossible to use YQL agent in clusters where both IPv4 and IPv6 are used simultaneously
(e.g. node domain names resovle in AAAA and kube-dns service domain names resolve in A).


savnadya:
Shipped!

PR #59, opened 2023-11-27
Call master_exit_read_only in master initJob. Fix #57

### LINKED ISSUE #57: Support master_exit_read_only in Kubernetes operator
Since version 23.2 read_only flag in master's Hydra is persistent, so after update `master_exit_read_only` is required.

l0kix2:

### DIFF
--- a/pkg/components/master.go
+++ b/pkg/components/master.go
@@ -194,8 +194,10 @@ func (m *master) createInitScript() string {
 func (m *master) createExitReadOnlyScript() string {
 	script := []string{
 		initJobWithNativeDriverPrologue(),
+		"export YT_LOG_LEVEL=DEBUG",
 		// COMPAT(l0kix2): remove || part when the compatibility with 23.1 and older is dropped.
-		"/usr/bin/yt execute master_exit_read_only '{}' || echo 'master_exit_read_only is supported since 23.2'",
+		`[[ "$YTSAURUS_VERSION" < "23.2" ]] && echo "master_exit_read_only is supported since 23.2, nothing to do" && exit 0`,
+		"/usr/bin/yt execute master_exit_read_only '{}'",
 	}
 
 	return strings.Join(script, "\n")
--- a/pkg/components/init_job.go
+++ b/pkg/components/init_job.go
@@ -9,14 +9,15 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
 	"github.com/ytsaurus/yt-k8s-operator/pkg/apiproxy"
 	"github.com/ytsaurus/yt-k8s-operator/pkg/consts"
 	"github.com/ytsaurus/yt-k8s-operator/pkg/labeller"
 	"github.com/ytsaurus/yt-k8s-operator/pkg/resources"
 	"github.com/ytsaurus/yt-k8s-operator/pkg/ytconfig"
-	batchv1 "k8s.io/api/batch/v1"
-	corev1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const initJobPrologue = `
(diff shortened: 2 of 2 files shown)


pkg/components/master.go
```diff
@@ -159,6 +159,7 @@ func (m *master) createInitScript() string {
 
 	script := []string{
 		initJobWithNativeDriverPrologue(),
+		"/usr/bin/yt execute master_exit_read_only '{}'",
```
zlobober, line 162: Wouldn't it break working with old images, if this command fails with "command maxter_exit_read_only not found"? Should we add "|| true" here?

zlobober (approved):

pkg/components/master.go
```diff
@@ -159,6 +159,7 @@ func (m *master) createInitScript() string {
 
 	script := []string{
 		initJobWithNativeDriverPrologue(),
+		"/usr/bin/yt execute master_exit_read_only '{}' || echo 'master_exit_read_only is supported since 23.2'",
```
gritukan, line 162: Забавное наблюдение: этот вызов может пофейлиться не только из-за неподдержанного вызова, но и по случайным причинам. Но даже в таком случае упадёт следующий (мутирующий) вызов remove. Тем не менее предлагаю написать комментарий и оставить тудушку убрать || через некоторое время.

l0kix2: я подумал детектить конкретную ошибку будет оверкилл
давай напишу тудушку

pkg/components/master.go
```diff
@@ -159,6 +159,8 @@ func (m *master) createInitScript() string {
 
 	script := []string{
 		initJobWithNativeDriverPrologue(),
+		// TODO: remove || in the future
+		"/usr/bin/yt execute master_exit_read_only '{}' || echo 'master_exit_read_only is supported since 23.2'",
```
savnadya, line 163: It will not work. Init job runs only during cluster initialization, and it won't run again during update. You need to create one more job, which will be run after `UpdateStatusWaitingForPodsCreation`.

pkg/components/master.go
```diff
@@ -174,6 +191,16 @@ func (m *master) createInitScript() string {
 	return strings.Join(script, "\n")
 }
 
+func (m *master) createExitReadOnlyScript() string {
+	script := []string{
+		initJobWithNativeDriverPrologue(),
+		// TODO: remove || in the future
+		"/usr/bin/yt execute master_exit_read_only '{}' || echo 'master_exit_read_only is supported since 23.2'",
```
zlobober, line 198: We have a nice conventions for the compatibility code that must be erased in some future. It must be prepended with a comment of form `COMPAT(username): remove something when something happens`. The important part is to specify explicitly, what condition must hold in order to remove this COMPAT; this allows code maintainers to periodically grep codebase for the word "COMPAT" and easily understand, which COMPATs may be safely removed without trying to remember the details of the code 6 months ago.

In your case I'd write
```
// COMPAT(l0kix2): remove || part when the compatibility with 23.1 and older is dropped.
```
(also, mind the dot in the end of the line)

pkg/components/master.go
```diff
@@ -218,3 +250,50 @@ func (m *master) Sync(ctx context.Context) error {
...
+		status, err := m.exitReadOnlyJob.Sync(ctx, dry)
+		return &status, err
+	}
+
+	if !dry {
+		m.ytsaurus.SetUpdateStatusCondition(metav1.Condition{
+			Type:    consts.ConditionMasterExitedReadOnly,
+			Status:  metav1.ConditionTrue,
+			Reason:  "MasterExitedReadOnly",
+			Message: "Master exited read only",
```
zlobober, line 285: Master exited read-only state

zlobober (approved):
LGTM modulo cosmetics

l0kix2:
I still need to run some tests before merge anyway. I've ran the case when we deploy ytsaurus cluster and then update base image and everything went as expected. But I'm not sure yet about the case when we deploy cluster with vversion A, then update it to version B and then update to version C (that time need to make sure that existing job will be correctly restarted), but for now i've had some problems, because some of init scripts hanged and complains about master unavailability and I've not yet localized is problem with my local setup or ytsaurus image version or the new code.

l0kix2:
- Fixed codestyle notes
- Successively ran e2e tests
- Tested new exit read only job switching from `ytsaurus/ytsaurus:stable-23.1.0` to `ytsaurus/ytsaurus:stable-23.1.0-relwithdebinfo` and back (script just gracefully fails for that version since cmd is not supported, though job workflow looks fine).

I'll try to test against 23.2 if I be able to find stable enough images.

l0kix2:
- I've tested operator case "deploy verA, update to ver B, update to ver C" against both 23.1 and 23.2 ytsaurus versions and it works as expected.
- Also added check for YTsaurus version in `master_exit_read_only` job script, so now we either do nothing for versions <=23.1 and or run `master_exit_read_only` cmd otherwise (instead of suppressing possible errors).

PR #68, opened 2023-12-07
e2e test for 23.1 > 23.2 update

l0kix2:

Add e2e to check that updating from 23.1 to 23.2 (and master_exit_read_only job) works.

### DIFF
--- a/controllers/ytsaurus_controller_test.go
+++ b/controllers/ytsaurus_controller_test.go
@@ -173,73 +173,20 @@ type testRow struct {
 
 var _ = Describe("Basic test for Ytsaurus controller", func() {
 	Context("When setting up the test environment", func() {
-		It("Should run and update Ytsaurus", func() {
-
-			By("Creating a Ytsaurus resource")
-			ctx := context.Background()
-
-			namespace := "test1"
-
-			ytsaurus := ytv1.CreateBaseYtsaurusResource(namespace)
-
-			g := ytconfig.NewGenerator(ytsaurus, "local")
-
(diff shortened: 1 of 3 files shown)


controllers/ytsaurus_controller_test.go
```diff
@@ -173,73 +173,20 @@ type testRow struct {
...
-				}
-				return ytsaurus.Status.State == ytv1.ClusterStateRunning
-			}, timeout*5, interval).Should(BeTrue())
-
-			By("Check that cluster alive after update")
-
-			Expect(ytClient.ListNode(ctx, ypath.Path("/"), &res, nil)).Should(Succeed())
-		})
+		It(
+			"Should run and update Ytsaurus to the same minor version",
```
savnadya, line 177: major

achulkov2: In terms of YT all numbered versions are historically called "major" releases, and we use the term minor for updates within a single major version, i.e. from some 23.1 commit to a later 23.1 commit.

l0kix2: So, if I've correctly understood from your comment, 23.1 and 23.2 are called major version and it would be correct to write something like:
`Should run and update Ytsaurus within same major version` for 23.1>23.1 case and 
`Should run and update Ytsaurus to the next major version` for 23.1>23.2 case.
?

l0kix2: Fixed

savnadya (approved):

savnadya:
Shipped!

PR #69, opened 2023-12-08
Add access control object namespace "queries" and object "nobody" during init

Krock21rus:
We implemented server side changes according to https://github.com/ytsaurus/ytsaurus/issues/175 proposal

They assume existence of "queries" access control object namespace and "nobody" access control object in a cluster.

Adding creation of these objects during query tracker init job


Krock21rus:
Added a test

savnadya:
Shipped!

PR #70, opened 2023-12-08
Fixed masters hostnames

### LINKED ISSUE #60: Fixed host addresses for masters
Let us recall that each pod has two addresses: pod address (which looks like `ms-2.masters.<namespace>.svc.cluster.local`) and host address, which is an FQDN of underlying k8s node.

Problem: in hostNetwork = true setting we are exposing master in a very hacky way. 
1. We specify hostNetwork = true for pods, making server processes bind to host network interfaces (making them available from outside k8s cluster). Actually, all our server processes bind to wildcard network interfaces, so masters even bind to CNI network interfaces too, allowing accessing them from their pod domain names.
2. We still use pod names in cluster_connection section of other components, as well as in master configs describing master cells; this work because of the remark from p.1
3. In order for master to identify himself in the list of peers int he cell, it searches for his fqdn in the list of peers. This does not work well because fqdn resolves to a host address, and the list of peers contains pod addresses, so we use an address resolver hack: we set /address_resolver/localhost_name_override to master's pod address with env variable substitution. This makes master being able to start and work.

This approach has a lot of downsides.
1. We are mixing pod addresses and host addresses for different components; in case when k8s overlay network uses ipv4 and host network uses ipv6, this requires setting up dualstack address resoltuion in all components, which is in general a bad practice.
2. We are relying on KubeDNS for pod addresses, which has a number of disadvantages (most notorious of which is c-ares DNS spoofing issue, which sometimes breaks DNS resolution at all). 
3. Masters register under pod addresses in //sys/primary_masters, which makes it hard to find their host addresses from outside the cluster (still possible via annotations, but not convenient at all). This complicates integration with outside tools like pull-based monitoring.
4. Finally, the incorrect addresses of masters get into cluster connection, which makes it impossible to connect components outside of the cluster using the config `//sys/@cluster_connection`.

The proposal is to switch to fixing k8s nodes for masters a priori by their fqdns. If we know host fqdns of these nodes, a user must specify it as follows:

```
masterHostAddresses:
- 4932:
  - host1.external.address
  - host2.external.address
  - host3.external.address
```

Here 4932 is the cell tag of a master cell (for potential extension to the multicell case). 

This option should be specified if and only if hostNetwork: true holds. In such case three changes in logic must apply.
1. Specified host addresses must be put in master configs.
2. Specified host addresses must be put in cluster connection configs of all components.
3. Master statefulset must get an additional affinity constraint of form externalHostnameLabel IN (host1.external.address, host2.external.address, host3.external.address), where externalHostnameLabel is `kubernetes.io/hostname` by default, but may be overridden in case when the value of kubernetes.io/hostname does not resolve outside, but some other label contains actual external addresses of nodes.

Also, the address_resolver hack must be dropped unconditionally. It is ugly.

l0kix2:

Added ability to bind YTsaurus masters to the set of nodes by node hostnames. 

 - New `HostAddresses ` field can be set only if `hostNetwork=true`;
 - values from `HostAddresses ` are used in all components' configs for connecting to the masters.
 - masters' statefulset has node affinity to deploy pods on the desired set of nodes.
 - `LocalhostNameOverride` is ignored if new `HostAddresses` is being used.

UPD. I forgot to use login in branchname, sorry

Other:
 - LocalhostNameOverride hack is deleted in separate PR https://github.com/ytsaurus/yt-k8s-operator/pull/74 which must be merged some time later — so we can gracefully update operator first, switch from `LocalhostNameOverride` to `HostAddresses` everywhere and only then update to new version of operator without LocalhostNameOverride

Fixes #60


pkg/components/master.go
```diff
@@ -253,6 +258,51 @@ func (m *master) Sync(ctx context.Context) error {
...
+	if affinity.NodeAffinity != nil {
+		nodeAffinity = affinity.NodeAffinity
+	}
+
+	selector := &corev1.NodeSelector{}
+	if nodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution != nil {
+		selector = nodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution
+	}
+
+	nodeHostnameLabel := primaryMastersSpec.HostAddressLabel
```
zlobober, line 288: I'd suggest introducing helper like m.HostAddressLbaelOrDefault()

pkg/ytconfig/generator.go
```diff
@@ -282,7 +292,7 @@ func (g *Generator) getMasterConfigImpl() (MasterServer, error) {
 	g.fillPrimaryMaster(&c.PrimaryMaster)
 	configureMasterServerCypressManager(g.ytsaurus.Spec, &c.CypressManager)
 
-	if g.ytsaurus.Spec.HostNetwork {
+	if g.ytsaurus.Spec.HostNetwork && len(spec.HostAddresses) == 0 {
 		// Each master deduces its index within cell by looking up his FQDN in the
```
zlobober, line 296: Now that this logic becomes legacy, let's add a comment of form:
```
// COMPAT(lokix2): remove that after we drop support for specifying host network without master host addresses.
```

zlobober (approved):

l0kix2:
Fixed review comments.

savnadya:
Shipped!

savnadya:
Shipped!

api/v1/ytsaurus_webhook.go
```diff
@@ -119,6 +120,35 @@ func (r *Ytsaurus) validateSecondaryMasters(old *runtime.Object) field.ErrorList
...
+		)
+	}
+
+	if len(masterSpec.HostAddresses) != 0 && len(masterSpec.HostAddresses) != int(masterSpec.InstanceCount) {
+		allErrors = append(
+			allErrors,
+			field.Invalid(
+				hostAddressesFieldPath,
+				masterSpec.HostAddresses,
+				fmt.Sprintf("%s list length shoud be equal to instanceCount", hostAddressesFieldPath.String()),
```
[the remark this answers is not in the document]
l0kix2: Now I see that validation is a bit wrong. We should only check if any (or even only one of) of `HostAddresses` or `InstanceCount` is set, because `InstanceCount` makes no sense in that case and not used.

PR #77, opened 2023-12-18
Set privileged for exec-node init containers too

koct9i:
Volume mount with mountPropagation = Bidirectional requires privileged
container. So it must be set also for init containers which have
the same set of volume mounts.

MountPropagation = Bidirectional is required for propagating tmpfs
mounts from exec-node namespace into host and then into containerd
sidecar namespace. For now this is simplest solution to get tmpfs.


savnadya:
Shipped!

Issue #75, opened 2023-12-13
Remote node CRD


zlobober:
We'd like to have an opportunity to describe a group of nodes (dat/exe/tab) as a separate CRD, providing the endpoints of a remote cluster. Such group of nodes (called remote nodes) will work the same as if they were described within the ytsaurus CRD, but they can belong to a different k8s cluster, which may be useful.

I'd suggest having a CRD RemoteNodeGroup, which looks exactly the same as dataNodes/execNodes/tabletNodes (and, ideally, reuses as much of existing code as possible). 

The main difference is that there should be a field "remoteClusterSpec" which must be a reference to a resource (CRD) of type RemoteYtsaurusSpec.

RemoteYtsaurusSpec must describe necessary information for connecting a node to a remote cluster. remoteClusterSpec must be a structure, currently containing one field:
```
MasterAddresses: []string
```

But in the future it will probably be extended by various other fields.
When https://github.com/ytsaurus/ytsaurus/issues/248 completes:
```
RemoteClusterUrl: string // allows taking cluster connection by HTTP cal to RemoteClusterUrl + "/cluster_connection".
```

When authentication in native protocol is employed:
```
Credentials: SomeKindOfCredentials
```

psushin:
Should RemoteYsaurusSpec be a separate CRD, or just described in a config map?
This CRD doesn't have any behavior, as far as I can see, so no particular controller, reconciliation, etc.

zlobober:
We discussed this matter with k8s-experienced colleagues, and it seems like having CRD for such case is a common practice. Configmap is unschematized object, and CRD is schematized, which is much less error-prone. In general, CRD does not necessarily require a designated controller, it may be a part of a more complex CRD like in our case.

l0kix2:
This change will be separated in a chain of PRs to ease codereviewing.
0. (merged) #83 in which we increase test coverage of config generator so we don't break things in following refactoring;
1. #84 in which config generator is being refactored to be able to work with remote nodes in the future.
2. (WIP) server/statefulset refactoring to be able to work without full ytsaurus spec with only master connection
3. (WIP) new CRDs for remote nodes & remote ytsaurus + reconcillers

l0kix2:
> This CRD doesn't have any behavior, as far as I can see, so no particular controller, reconciliation, etc.

> In general, CRD does not necessarily require a designated controller, it may be a part of a more complex CRD like in our case.

[...]

PR #83, opened 2023-12-27
Config generator tests for all components

l0kix2:

First PR for #75

Currently we have test coverage for YTsaurus config generating code only for master and scheduler. Here I added tests for other components, so config generator can be easily refactored if needed.

Changes are only in test code and existing tests behaviour is not changed.


pkg/ytconfig/generator_test.go
```diff
@@ -14,61 +15,180 @@ import (
...
+		Rack: "fake",
+	}
 )
 
+func TestGetChytInitClusterConfig(t *testing.T) {
+	g := NewGenerator(getYtsaurus(), testClusterDomain)
+	cfg, err := g.GetChytInitClusterConfig()
+	require.NoError(t, err)
+	canonize.Assert(t, cfg)
+}
```
savnadya, line 75: Let's add empty lines between functions

pkg/ytconfig/generator_test.go
```diff
@@ -14,61 +15,180 @@ import (
...
+func TestGetDiscoveryConfig(t *testing.T) {
+	g := NewGenerator(withDiscovery(getYtsaurus()), testClusterDomain)
+	cfg, err := g.GetDiscoveryConfig()
+	require.NoError(t, err)
+	canonize.Assert(t, cfg)
+}
+func TestGetExecNodeConfig(t *testing.T) {
+	g := NewGenerator(getYtsaurus(), testClusterDomain)
```
savnadya, line 95: Maybe add exec nodes in ytsaurus spec?
Same for data nodes

savnadya (changes requested):

pkg/ytconfig/generator_test.go
l0kix2: As we agreed in chat, I've introduced a new function "ytsaurus spec with every component" and used it in all non-specific tests, so we test generating configs for more realistic spec. 
And for tests that want to test something specific (like master hostnames) we can build spec granularly adding required functions.

savnadya:
Shipped!

PR #87, opened 2023-12-29
[remote nodes] 2. Refactor statefulset & server

l0kix2:
Third PR for #75.

**The issue:**
Currently StatefulSet and server code can work only if they have full ytsaurus CRD spec. But for remote nodes feature we need a way to create data/tablet/exec nodes servers with stateful sets without having ytsaurus spec. 

**What is done:**
This is refactoring without changing current behaviour (do we have unit tests here or only e2e?)

1. For `StatefulSet` and `serverImpl` instead of passing `apiproxy.Ytsaurus` reference we pass `apiproxy.APIProxy` + `ytv1.CommonSpec` (which was introduced in #84). It is enough for StatefulSet code, though some fields are moved to `CommonSpec` from `YtsaurusSpec`.
2. Instead of single `componentBase` I've introduced bunch of structs to reuse some of the behaviour: 
   - `labellable` (lableller): only `InitJob` now
and couple of presets
   - `localComponent` (`labellable` + `apiproxy.Ytsaurus`): for `strawberryController`, `UI`, `ytsaurusClient`
   - `localServerComponent` (`localComponent` + `server`): for almost every other component we have.
This allowed to remove `server` dependency on checking if `ytsaurus` is updating, therefore giving the opportunity to use server for remote nodes codes (unaware of `ytsaurus` component).

I wouldn't say this hierarchy looks perfect semantically, though IMO this change more or less easy to understand in the current codebase.


l0kix2:
@savnadya I've removed the `backlog` label here, because it is the most in progress task I'm occupied right now. Hope the removing doesn't break something.

pkg/components/component.go
```diff
@@ -54,20 +54,39 @@ type Component interface {
 	IsUpdatable() bool
 }
 
-type componentBase struct {
+type labellable struct {
```
zlobober, line 57: Let's use this as a cause to write a couple of comments explaining the meaning of this struct, and what are Name and Label. It is hard to recall their difference for me.

pkg/components/component.go
```diff
@@ -54,20 +54,39 @@ type Component interface {
...
 	return c.labeller.ComponentName
 }
 
-func (c *componentBase) GetLabel() string {
+func (c *labellable) GetLabel() string {
 	return c.labeller.ComponentLabel
 }
 
-func (c *componentBase) SetReadyCondition(status ComponentStatus) {
+type localComponent struct {
```
zlobober, line 69: Also, a couple of comments (a-la the ones from the PR description) would look nice here.

zlobober (approved):
LGTM modulo my request about writing a couple more comments.

pkg/components/component.go
l0kix2: Fun thing is it seems that GetLabel was not used at all. So I deleted it.

pkg/components/component.go
l0kix2: done

PR #92, opened 2024-01-08
Support local k8s port forwarding in E2E

l0kix2:

Currently E2E tests are only work on linux. It is impossible (or I couldn't find a way) to access internal container port (yt http proxy being specifically) from localhost which is not exposed. Though it is possible for mac with port-forwarding. 

So here I've added possibility to specify internal port for http proxy pod in `E2E_HTTP_PROXY_INTERNAL_PORT` env var and localhost yt proxy using `E2E_YT_PROXY` env var. That way if env vars configured consistently with port forwarding test can be run on mac os.

Current tests behaviour is left unchanged.

### DIFF
--- a/api/v1/test_helpers.go
+++ b/api/v1/test_helpers.go
@@ -1,6 +1,9 @@
 package v1
 
 import (
+	"os"
+	"strconv"
+
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


savnadya:
Shipped!

PR #97, opened 2024-01-09
Enlarge Job backoff limit

l0kix2:

Default is 6 and it isn't much on practice.


savnadya:
Shipped!

PR #98, opened 2024-01-10
Add set condition logging

l0kix2:

For better troubleshooting.


savnadya:
Shipped!

PR #99, opened 2024-01-10
Fix master update state

l0kix2:

repro is here https://github.com/ytsaurus/yt-k8s-operator/pull/95


savnadya:
Shipped!

PR #101, opened 2024-01-10
Fix master schema ACLs and improve init jobs

achulkov2:
This PR makes several improvements:
* The operator now creates a group for admins (`admins`).
* The operator now sets better ACLs for schemas.
* Master and scheduler init jobs do not overwrite existing dynamic configs anymore.

### DIFF
--- a/pkg/components/helpers.go
+++ b/pkg/components/helpers.go
@@ -7,9 +7,11 @@ import (
 	"github.com/ytsaurus/yt-k8s-operator/pkg/apiproxy"
 	"go.ytsaurus.tech/library/go/ptr"
 	"go.ytsaurus.tech/yt/go/ypath"
+	"go.ytsaurus.tech/yt/go/yson"
 	"go.ytsaurus.tech/yt/go/yt"
 	"k8s.io/utils/strings/slices"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"strings"
 )
 
 func CreateTabletCells(ctx context.Context, ytClient yt.Client, bundle string, tabletCellCount int) error {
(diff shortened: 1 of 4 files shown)


pkg/components/master.go
```diff
@@ -173,6 +174,77 @@ func (m *master) initMedia() string {
...
+		SetPathAcl("//sys/schemas/medium", []yt.ACE{
+			userReadACE,
+			adminACE,
+		}),
+		SetPathAcl("//sys/schemas/domestic_medium", []yt.ACE{
+			userReadCreateACE,
+			adminACE,
+		})))
+
+	// Users can to create pools, pool trees and accounts given the right circumstances and permissions.
```
zlobober, line 231: "to" is excessive

pkg/components/master.go
```diff
@@ -173,6 +174,77 @@ func (m *master) initMedia() string {
...
+	}
+
+	var commands []string
+
+	// Users should not be able to create or write objects of these types on their own.
+	for _, objectType := range []string{
+		"tablet_cell", "tablet_action", "tablet_cell_bundle",
+		"user", "group",
+		"rack", "data_center", "cluster_node",
+		"access_control_object", "access_control_object_namespace", "access_control_object_namespace_map"} {
```
Krock21, line 214: We may want to allow users to create access control objects (chyt cliques access control, queries access control)

We may want to add `{"allow" "owner" "read,write,administer"}` to access_control_object_namespace so newly created access control objects will be accessible by creators

achulkov2: When would something like this be needed? For CHYT, the strawberry controller is responsible for creating the aco namespace and object, and automatically sets permission for the user who created the clique. 
What will the pipeline look like for QT access control?

l0kix2 (approved):
LGTM

pkg/components/helpers.go
```diff
@@ -135,3 +137,31 @@ func handleUpdatingClusterState(
 	}
 	return nil, err
 }
+
+func SetPathAcl(path string, acl []yt.ACE) string {
+	aclAsString, err := yson.MarshalFormat(acl, yson.FormatText)
```
l0kix2, line 142: I suppose this shouldn't be called `*asString` since it is bytes.

pkg/components/master.go
Krock21: For QT we expected users to create ACOs for themselves and we are developing UI for it

1. A user goes to Navigation -> //sys/access_control_object_namespaces/queries and clicks "Create Access Control Object"
2. A user specifies ACL for ACO
3. A user goes to Queries, writes a query, and selects newly created ACO from a list to guard a query

So this is needed already

> We may want to add `{"allow" "owner" "read,write,administer"}` to access_control_object_namespace

So, current structure is:
1. access_control_object_namespace schema specifies rules for all namespaces
2. "chyt" and "queries" namespaces inherit acl from schema
3. ACOs inherit acls from their namespaces

I think the best solution would be:
1. `access_control_object` schema will be `{"allow" "users" "create"}`. We don't need "read" since only create matters([since ACOs inherit from their namespace, not schema](https://github.com/ytsaurus/ytsaurus/blob/717e7109204b42120c212be427a2de75159cd7f5/yt/yt/server/master/cypress_server/access_control_object_type_handler.cpp#L112)). We may keep "read" for simplicity
2. `access_control_object_namespace` will be `{"allow" "users" "read"}`
3. namespace "chyt" untouched
4. namespace "queries" adds in their ACL `{"allow" "users" "create_children"}`
    + I don't see in documentation something like "create_children", but with experimenting I found that creation requires `"write | modify_children"` permission. Please correct me here on what permission should we use

achulkov2: Okay, so it seems like the user only needs to create ACOs, not ACO namespaces?

So in terms of this PR, I will modify the access_control_object schema as stated in 1, 2 is done already, 4 is something that should be done by QT/its init job. 

P.S. write | modify_children seems correct, that's how it works for scheduler pools as well

Krock21: >the user only needs to create ACOs, not ACO namespaces?
Right

>4 is something that should be done by QT/its init job.
Agreed

pkg/components/helpers.go
achulkov2: Ok, I will rename it to `formattedAcl`.

savnadya:
Shipped!

savnadya (approved):

PR #103, opened 2024-01-11
Fix flapping test

l0kix2:

The bug is not really easy to test so test is flapping.
Originally idea was that we wait till `PossibilityCheck` after the update and check if master was recreated right after. Though with bug fix I suppose cluster just goes to next update states very fast and test never sees `PossibilityCheck` status and fails.

I've added a artificial condition for deploy to stay in `PossibilityCheck` condition: I've created duplicate of leader in //sys/primary_master, so operator will indefinitely stay in `PossibilityCheck` loop (btw that's what is really happened with real cluster i've been updating recently). So now test shouldn't flap I hope.


savnadya:
Shipped!

PR #105, opened 2024-01-17
CRI containerd job environment

koct9i:
- Cleanup handling legacy config layout
- Cleanup building spec for exec nodes
- Add jobResources and jobEnvironment into exec node spec
- Ignore typed nil in Fetch And Sync
- Add CRI job environment
- Add EntrypointWrapper into InstanceSpec


pkg/ytconfig/node.go
```diff
@@ -317,7 +362,7 @@ func getExecNodeLogging(spec *ytv1.ExecNodesSpec) Logging {
 		[]ytv1.TextLoggerSpec{defaultInfoLoggerSpec(), defaultStderrLoggerSpec()})
 }
 
-func getExecNodeServerCarcass(spec *ytv1.ExecNodesSpec, usePorto bool) (ExecNodeServer, error) {
+func getExecNodeServerCarcass(spec *ytv1.ExecNodesSpec, ytsaurusSpec *ytv1.YtsaurusSpec) (ExecNodeServer, error) {
```
l0kix2, line 365: This change would conflict with #75 feature, in which i've made (not yet merged) some refactorings to be able to use code (in particular config generator) without having `YtsaurusSpec` (because remote nodes won't have one). 
Could you please not add ytsaurusSpec parameter and just pass needed arguments as is (and maybe in future they can be combine them in some struct if needed).

l0kix2 (changes requested):
I can't say much about containerd stuff here, suppose it's ok, but I suggest to add some testing (either e2e or envtest-based without real cluster — I hope some helpers for second option will be merged in main soon) for this new functionality.

Also I suppose we better test some existing exec nodes code current behaviour, it is not obviously to me we don't break anything from diff.

pkg/ytconfig/cri.go
```diff
@@ -0,0 +1,67 @@
+package ytconfig
```
l0kix2, line 1: Let's add canonized test for exec node spec with new fields. It should be easy since we have a lot of them.

l0kix2 (changes requested):

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/ytsaurus/yt-k8s-operator
 go 1.21
 
 require (
+	github.com/BurntSushi/toml v1.3.2
```
gritukan, line 6: Nice login :)

pkg/components/exec_node_base.go
```diff
@@ -3,56 +3,147 @@ package components
...
-	sidecars   []string
-	privileged bool
+	server server
+	cfgen  *ytconfig.NodeGenerator
+	spec   *ytv1.ExecNodesSpec
+
+	sidecarConfig *ConfigHelper
+}
+
+func (n *baseExecNode) IsJobResourcesIsolated() bool {
```
gritukan, line 27: AreJobResourcesIsolated? Or maybe even better AreJobsIsolated

pkg/components/exec_node_base.go
```diff
@@ -3,56 +3,147 @@ package components
...
+	if n.spec.JobEnvironment != nil && n.spec.JobEnvironment.CRI != nil {
```
gritukan, line 28: Just `return n.spec.JobEnvironment != nil && n.spec.JobEnvironment.CRI != nil`?

pkg/components/exec_node_base.go
```diff
@@ -3,56 +3,147 @@ package components
...
-				ct.SecurityContext = &corev1.SecurityContext{}
+func (n *baseExecNode) doBuildBase() error {
+	statefulSet := n.server.buildStatefulSet()
+	podSpec := &statefulSet.Spec.Template.Spec
+
+	// Pour job resources into node container if jobs are not isolated.
+	if n.spec.JobResources != nil && !n.IsJobResourcesIsolated() {
+		addResourceList := func(list, newList corev1.ResourceList) {
+			for name, quantity := range newList {
+				if value, ok := list[name]; !ok {
```
gritukan, line 46: Maybe invert if?

pkg/components/exec_node_base.go
```diff
@@ -3,56 +3,147 @@ package components
...
+
+	setContainerPrivileged := func(ct *corev1.Container) {
+		if ct.SecurityContext == nil {
+			ct.SecurityContext = &corev1.SecurityContext{}
+		}
+		ct.SecurityContext.Privileged = ptr.Bool(n.spec.Privileged)
+	}
+
+	if len(podSpec.Containers) != 1 {
+		log.Panicf("length of exec node containers is expected to be 1, actual %v", len(podSpec.Containers))
```
gritukan, line 67: Number of

pkg/ytconfig/cri.go
```diff
@@ -0,0 +1,67 @@
...
+
+				"cni": map[string]any{
+					"conf_dir": "/etc/cni/net.d",
+					"bin_dir":  "/usr/local/lib/cni",
+				},
+
+				"containerd": map[string]any{
+					"default_runtime_name": "runc",
+					"runtimes": map[string]any{
+						"runc": map[string]any{
```
gritukan, line 52: Is it true that this config will be stored into the file and it will be possible to add an init container that executes `nvidia-container-toolkit` which adds nvidia runtime to the config?

pkg/components/exec_node_base.go
koct9i: I've ended with IsJobEnvironmentIsolated

pkg/components/exec_node_base.go
koct9i: Now this piece is more complicated.

pkg/components/exec_node_base.go
koct9i: Done. Not my code, I've just scrambled it. =)

pkg/ytconfig/cri.go
koct9i: Yes. I've added "entrypointWrapper" which could do this and much more.

api/v1/ytsaurus_types.go
```diff
@@ -342,6 +348,42 @@ type DataNodesSpec struct {
...
+	// CRI namespace for jobs.
+	//+optional
+	Namespace *string `json:"namespace,omitempty"`
+	// Base cgroup for jobs.
+	//+optional
+	BaseCgroup *string `json:"baseCgroup,omitempty"`
+}
+
+type JobEnvironmentSpec struct {
+	// Isolate job execution environment from exec node or not, by default true when possible.
```
gritukan, line 370: Maybe separate?

api/v1/ytsaurus_types.go
```diff
@@ -342,6 +348,42 @@ type DataNodesSpec struct {
...
+type CRIJobEnvironmentSpec struct {
+	// Specifies wrapper for CRI service command.
+	//+optional
+	EntrypointWrapper []string `json:"entrypointWrapper,omitempty"`
+	// Sandbox (pause) image.
+	//+optional
+	SandboxImage *string `json:"sandboxImage,omitempty"`
+	// Timeout for retrying CRI API calls.
+	//+optional
+	RetryTimeoutSeconds *int32 `json:"retryTimeoutSeconds,omitempty"`
```
gritukan, line 360: I'd suggest APIRetryTimeoutSeconds

api/v1/ytsaurus_types.go
```diff
@@ -342,6 +348,42 @@ type DataNodesSpec struct {
...
+	// CRI namespace for jobs.
+	//+optional
+	Namespace *string `json:"namespace,omitempty"`
```
gritukan, line 363: Maybe CRINamespace?

api/v1/ytsaurus_types.go
koct9i: No, i like "isolation" here.

"Isolate" here means jobs are isolated from the exec node while node stays where it is. 
There are strong subordination between components, "separate" is something about dividing comparable entities.

api/v1/ytsaurus_types.go
```diff
@@ -342,6 +348,42 @@ type DataNodesSpec struct {
 	Name string `json:"name,omitempty"`
 }
 
+type CRIJobEnvironmentSpec struct {
+	// Specifies wrapper for CRI service command.
+	//+optional
+	EntrypointWrapper []string `json:"entrypointWrapper,omitempty"`
```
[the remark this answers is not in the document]
koct9i: I've called here it "CRI service"

l0kix2 (approved):
lgtm, though suppose e2e which would successfully run job, relying on some specific image env would help not to break this code in the future.

l0kix2 (changes requested):
Tests are failing though

l0kix2 (approved):

PR #108, opened 2024-01-22
feat: set layer locations for exec nodes

### LINKED ISSUE #107: Add a spec option to set layer locations for exec nodes

tdakkota:
Fixes #107.

### DIFF
--- a/api/v1/ytsaurus_types.go
+++ b/api/v1/ytsaurus_types.go
@@ -72,6 +72,7 @@ type LocationType string
 const (
 	LocationTypeChunkStore       LocationType = "ChunkStore"
 	LocationTypeChunkCache       LocationType = "ChunkCache"
+	LocationTypeChunkLayer       LocationType = "ChunkLayer"
 	LocationTypeSlots            LocationType = "Slots"
 	LocationTypeLogs             LocationType = "Logs"
 	LocationTypeMasterChangelogs LocationType = "MasterChangelogs"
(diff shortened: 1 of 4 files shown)


api/v1/ytsaurus_types.go
```diff
@@ -72,6 +72,7 @@ type LocationType string
 const (
 	LocationTypeChunkStore       LocationType = "ChunkStore"
 	LocationTypeChunkCache       LocationType = "ChunkCache"
+	LocationTypeChunkLayer       LocationType = "ChunkLayer"
```
l0kix2, line 75: It seems that `chunk` part is makes no sense here. Suggesting to call this field `LocationTypeLayers` or even `LocationTypePortoLayers` and change const value accordingly.

pkg/ytconfig/node.go
```diff
@@ -386,6 +421,16 @@ func getExecNodeServerCarcass(spec *ytv1.ExecNodesSpec, usePorto bool) (ExecNode
 	c.JobResourceManager.ResourceLimits = c.ExecNode.JobController.ResourceLimitsLegacy
 	c.ExecNode.GpuManager = c.ExecNode.JobController.GpuManager
 
+	if usePorto {
+		for _, loc := range spec.Locations {
+			c.ExecNode.RootFSBinds = append(c.ExecNode.RootFSBinds, RootFSBind{
+				InternalPath: loc.Path,
+				ExternalPath: loc.Path,
+				ReadOnly:     false,
+			})
```
l0kix2, line 430: Could you please add some comment with explaining why rootfs binds configuration is being built that way.

l0kix2 (changes requested):

PR #111, opened 2024-02-02
[remote nodes] 3. Remote CRDs and controller

l0kix2:
In this change we introducing two new CRDs: RemoteYtsaurus and RemoteExecNodes. For the purposes of easing code reviewing in this PR i have only added new types and generated code. controller implementation is in the next PR (#112)


config/crd/kustomization.yaml
```diff
@@ -13,13 +15,17 @@ patchesStrategicMerge:
 - patches/webhook_in_ytsaurus.yaml
 - patches/webhook_in_spyts.yaml
 - patches/webhook_in_chyts.yaml
+#- patches/webhook_in_remoteytsauruses.yaml
```
zlobober, line 18: Why are these commented?

zlobober (approved):

config/crd/kustomization.yaml
l0kix2: Not sure, it is what kubebulder generator does. I suppose if we want to add webhooks for remote ytsaurus or remotenodes CRDs we will uncomment this.

PR #113, opened 2024-02-05
[remote nodes] 5. e2e test

l0kix2:


l0kix2:
Hmm, the e2e test is currently failing because couldn't find scheduler for MR operation, but it was green earlier. Wonder what has changed.

controllers/ytsaurus_controller_test.go
```diff
@@ -400,6 +459,66 @@ var _ = Describe("Basic test for Ytsaurus controller", func() {
...
+			runRemoteYtsaurus(remoteYtsaurus)
+
+			defer deleteRemoteExecNodes(ctx, remoteNodes)
+			runRemoteExecNodes(remoteNodes)
+
+			By("Creating ytsaurus client")
+			ytClient := getYtClient(g, namespace)
+
+			By("Running sort operation to ensure exec node works")
+			runAndCheckSortOperation(ytClient)
```
zlobober, line 520: Let's return operation from this call, and make sure that all jobs within operation were indeed run on remote nodes? This can be done by list_jobs(op.id) call, and maybe get_job(job.id) for each job.

zlobober (changes requested):
LGTM after improving the test.

controllers/ytsaurus_controller_test.go
l0kix2: For some reason `ytClient.ListJobs` returns empty list of jobs. Any ideas why that could happen? For now i've tried to pass `JobState: &yt.JobCompleted` but result is the same.

zlobober: Were there any jobs at all? Can you see them in scheduler logs?

l0kix2: Well, my table was sorted as far as i can see.
I'll check the logs.

zlobober: In order to debug something like that, it is convenient to insert sleep(10000) after operation completion and connect to cluster using yt CLI, try that also.

PR #116, opened 2024-02-08
Rename operations archive

omgronny:


pkg/components/scheduler.go
```diff
@@ -283,10 +283,18 @@ func (s *scheduler) createInitUserScript() string {
 	return strings.Join(script, "\n")
 }
 
-func (s *scheduler) prepareInitOperationArchive() {
+func (s *scheduler) prepareInitOperationsArchive() {
```
l0kix2, line 286: Could you please add comment explaining what is the reason for migration here (new script in some new ytsaurus version or same script was renamed or smth)?

l0kix2 (changes requested):

pkg/components/scheduler.go
```diff
@@ -283,10 +283,18 @@ func (s *scheduler) createInitUserScript() string {
...
+	initOpArchivePath = "/usr/bin/init_operations_archive"
+	// COMPAT(omgronny): drop once new srcipt name is everywhere
+	if _, err := os.Stat(initOpArchivePath); err != nil {
```
chizhonkova, line 289: Also you can reduce the number of ifs: `if _, err := os.Stat("/usr/bin/init_operations_archive"); errors.Is(err, os.ErrNotExist)`

l0kix2:
Also please check the test run https://github.com/ytsaurus/yt-k8s-operator/actions/runs/7829875497/job/21362727551?pr=116 — there are syntax errors.

pkg/components/scheduler.go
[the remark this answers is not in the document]
omgronny: What is the correct way to get the valid name then? I far as I know we can't just change it from `init_operation_archive` to `init_operations_archive` here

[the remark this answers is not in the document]
l0kix2: If you explain a little what change in ytsaurus you are trying to support in operator it would be easier to decide what to do, but I suppose is what you trying to achieve is to check file existence in the scheduler container, so you have to do that check _in the bash script_, which is being built below.

pkg/components/scheduler.go
l0kix2: As far as I see [currently](https://github.com/l0kix2/ytsaurus/blob/main/yt/docker/ytsaurus/Dockerfile#L26) script is called `/usr/bin/init_operation_archive` as in the operator's code. 
Do you plan to change the name of that script in the ytsaurus image in some version?

omgronny: [yes](https://github.com/ytsaurus/ytsaurus/commit/e5348fef221110ce27bba13df5f9790649084b01)

l0kix2: I see, then I suggest we write short comment that file was renamed in the image. Also it would be useful to provide this link to commit in the comment.

pkg/components/scheduler.go
```diff
@@ -283,10 +283,18 @@ func (s *scheduler) createInitUserScript() string {
...
+		if os.IsNotExist(err) {
+			initOpArchivePath = "/usr/bin/init_operation_archive"
+		}
+	}
 	script := []string{
 		initJobWithNativeDriverPrologue(),
-		fmt.Sprintf("/usr/bin/init_operation_archive --force --latest --proxy %s",
+		fmt.Sprintf("%s --force --latest --proxy %s",
```
savnadya, line 296: You can check init operation archive script path only inside this script

l0kix2 (changes requested):
Sorry that took so long, I missed that commits were added.

Looks good, could you please rebase on main branch and resolve conflict (I suppose it is `scheduler > Scheduler` rename)?

omgronny:
> Looks good, could you please rebase on main branch and resolve conflict (I suppose it is `scheduler > Scheduler` rename)?

Done

l0kix2:
one of e2es is failing, there is a chance that script is broken somehow

omgronny:
> one of e2es is failing, there is a chance that script is broken somehow

Actually, the tests were green until I accidentally closed and reopened PR. Is it possible we have some flaps?

l0kix2:
> > one of e2es is failing, there is a chance that script is broken somehow
> 
> Actually, the tests were green until I accidentally closed and reopened PR. Is it possible we have some flaps?

i've restarted them, let's see if the issue stays.

l0kix2:
It may be unrelated (though persistent) flap, will look into that and merge after investigation.

l0kix2 (approved):

PR #121, opened 2024-02-21
add startup and readiness probes to sts component specs

kruftik:


koct9i:
For the reference:
https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes

pkg/components/server.go
```diff
@@ -218,6 +228,27 @@ func (s *serverImpl) rebuildStatefulSet() *appsv1.StatefulSet {
 				Command:      []string{s.binaryPath, "--config", path.Join(consts.ConfigMountPoint, fileNames[0])},
 				VolumeMounts: volumeMounts,
 				Resources:    s.instanceSpec.Resources,
+				StartupProbe: &corev1.Probe{
```
koct9i, line 231: Not sure why we need startup probe here.

pkg/components/server.go
```diff
@@ -17,6 +18,15 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
+const (
+	initialProbesDelaySeconds = 90
```
koct9i, line 22: Initial delay isn't required either.

pkg/components/server.go
```diff
@@ -17,6 +18,15 @@ import (
...
+
+	probesTimeoutSeconds = 3
+	probesPeriodSeconds  = 15
+
+	readinessProbeHTTPPath = "/solomon/status"
```
l0kix2, line 27: solomon path shouldn't be hardcoded in the public code, it should be configured in spec or somehow else.

pkg/components/server.go
```diff
@@ -17,6 +18,15 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
+const (
+	initialProbesDelaySeconds = 90
+
+	probesTimeoutSeconds = 3
+	probesPeriodSeconds  = 15
```
l0kix2, line 25: I suppose these params also should be in spec and not hardcoded.

l0kix2 (changes requested):
I suggest creating some `ProbeSpec` struct in types and  fields of that type in the `InstanceSpec` struct which is reused for most of the components.
Not sure if fields should be named `StartupProbe` & `ReadinessProbe` though, better not clash with k8s names, maybe `StartupProbeTemplate` or smth would do?

pkg/components/server.go
zlobober: Is there any specific need to use solomon?
I think it is ok to use host:monitoring_port/orchid/service, it is just fine. And I do not see any issue to hardcode this particular address in the code, it is a detail of YT's implementation, which is shared by absolutely all C++ server components.

kruftik:
@koct9i @l0kix2 @zlobober pls take another look at the changes

l0kix2 (approved):
LGTM 
@koct9i @zlobober do you have more comments?

koct9i:
Ok, LGTM.

koct9i (approved):

PR #122, opened 2024-02-22
Added master caches support

qurname2:
To resolve issue https://github.com/ytsaurus/yt-k8s-operator/issues/94


pkg/ytconfig/names.go
```diff
@@ -188,3 +188,20 @@ func (g *Generator) FormatComponentStringWithDefault(base string, name string) s
 	}
 	return base
 }
+
+func (g *Generator) GetMasterCachesStatefulSetName() string {
+	return g.getName("ms-cache")
```
l0kix2, line 193: I suggest "msc" since others are rather short
```
yt                 hp                                                 147/155   95d
yt                 ms                                                 3/3       95d
yt                 qa                                                 3/3       95d
yt                 qt                                                 3/3       80d
yt                 rp                                                 147/155   95d
yt                 sch                                                3/3       95d
yt                 tnd                                                147/155   95d
yt                 yqla                                               3/3       80d
```

pkg/ytconfig/generator.go
```diff
@@ -690,3 +691,51 @@ func (g *Generator) GetDiscoveryConfig() ([]byte, error) {
...
+		return MasterCacheServer{}, err
+	}
+	g.fillCommonService(&c.CommonServer, &spec.InstanceSpec)
+	g.fillBusServer(&c.CommonServer, spec.NativeTransport)
+	c.Addresses = g.getMasterCachesAddresses()
+
+	return c, nil
+}
+
+func (g *Generator) GetMasterCachesConfig() ([]byte, error) {
```
l0kix2, line 708: Please add a test for that method. There are a lot of tests of generators in generator_test.go file.

pkg/components/master_caches.go
```diff
@@ -0,0 +1,161 @@
...
+	return err
+}
+
+func (mc *masterCache) doServerSync(ctx context.Context) error {
+	statefulSet := mc.server.buildStatefulSet()
+	mc.addAffinity(statefulSet)
+	return mc.server.Sync(ctx)
+}
+
+func (mc *masterCache) addAffinity(statefulSet *appsv1.StatefulSet) {
```
l0kix2, line 119: let's create function for adding affinity to stateful set in helpers.go (a maybe create affinity.go) for example and use it in master and mastercache components.

l0kix2:
I suppose that tests are failing because of yamls wasn't updated. I suppose make generate will fix that.
Also there is no way to understand why e2e are failing from gh actions log, so the only way is to run them locally and check operator logs.

l0kix2 (changes requested):

pkg/ytconfig/names.go
qurname2: Yep, agree

pkg/ytconfig/generator.go
qurname2: Done

api/v1/test_helpers.go
```diff
@@ -12,8 +12,8 @@ import (
 
 const (
 	YtsaurusName     = "test-ytsaurus"
-	CoreImageFirst   = "ytsaurus/ytsaurus-nightly:dev-23.1-5f8638fc66f6e59c7a06708ed508804986a6579f"
-	CoreImageSecond  = "ytsaurus/ytsaurus-nightly:dev-23.1-9779e0140ff73f5a786bd5362313ef9a74fcd0de"
+	CoreImageFirst   = "ytsaurus/ytsaurus-nightly:dev-23.1-28ccaedbf353b870bedafb6e881ecf386a0a3779"
+	CoreImageSecond  = "ytsaurus/ytsaurus-nightly:dev-23.1-28ccaedbf353b870bedafb6e881ecf386a0a3779"
```
l0kix2, line 16: I see 2 problems here:
1) you have replaced two different images with same one — therefore e2e test `test-minor-update` will test different scenario which wasn't intended
2) there is a `TEST_IMAGES` list in the Makefile which (unfortunately) duplicates list of images from the current file. Images being loaded in kind before e2e test. I assume your problem with test running locally and not failing at CI is because pods are downloading images which takes more than timeout. So you must update Makefile accordingly to this file.

While we catch that problem could you please add a comment here that this list should be in sync with one in the Makefile until we figure something better.

api/v1/test_helpers.go
```diff
@@ -225,6 +225,120 @@ func CreateBaseYtsaurusResource(namespace string) *Ytsaurus {
 	return ytsaurus
 }
 
+func CreateBaseYtsaurusResourceWithMasterCaches(namespace string) *Ytsaurus {
```
l0kix2, line 228: If only masterCaches are being added here, lets call CreateBaseYtsaurusResource and add mastercaches to spec without copy pasting its code.

l0kix2 (changes requested):

api/v1/test_helpers.go
qurname2: yes, will do it

Makefile
```diff
@@ -5,7 +5,7 @@ IMG ?= controller:latest
 ENVTEST_K8S_VERSION = 1.24.2
 
 OPERATOR_IMAGE = ytsaurus/k8s-operator
-OPERATOR_TAG = 0.0.0-alpha
+OPERATOR_TAG = master-caches-test-0.3
```
l0kix2, line 8: Assume this is temporary

Makefile
```diff
@@ -83,12 +83,12 @@ canonize: manifests generate fmt vet envtest ## Canonize tests.
 helm-kind-install: ## Install helm chart from sources in kind.
 	docker build -t ${OPERATOR_IMAGE}:${OPERATOR_TAG} .
 	kind load docker-image ${OPERATOR_IMAGE}:${OPERATOR_TAG}
-	helm install ytsaurus $(OPERATOR_CHART)
+	helm upgrade -i ytsaurus $(OPERATOR_CHART) --set controllerManager.manager.image.repository=${OPERATOR_IMAGE} --set controllerManager.manager.image.tag=${OPERATOR_TAG}
```
l0kix2, line 86: could explain the change (maybe add comment)

api/v1/test_helpers.go
l0kix2: this is yet to be done

api/v1/test_helpers.go
```diff
@@ -11,10 +11,12 @@ import (
 )
 
 const (
-	YtsaurusName     = "test-ytsaurus"
-	CoreImageFirst   = "ytsaurus/ytsaurus-nightly:dev-23.1-5f8638fc66f6e59c7a06708ed508804986a6579f"
+	YtsaurusName = "test-ytsaurus"
+	// Images should be in sync with TEST_IMAGES variable in Makefile
+	// todo: come up with a more elegant solution
+	CoreImageFirst   = "ytsaurus/ytsaurus-nightly:dev-23.1-28ccaedbf353b870bedafb6e881ecf386a0a3779"
 	CoreImageSecond  = "ytsaurus/ytsaurus-nightly:dev-23.1-9779e0140ff73f5a786bd5362313ef9a74fcd0de"
```
l0kix2, line 18: I suppose CoreImageSecond commit should be later than CoreImageFirst (we don't guarantee update with version downgrade), is it satisfied now?

(though I'm not 100% percent sure that it is requirement inside same major version, @achulkov2 can you confirm?)

pkg/ytconfig/base_generator.go
```diff
@@ -14,19 +14,22 @@ type BaseGenerator struct {
...
 	discoveryInstanceCount int32
+	masterCachesSpec       *ytv1.MasterCachesSpec
 }
 
 func NewRemoteBaseGenerator(
 	key types.NamespacedName,
 	clusterDomain string,
 	commonSpec ytv1.CommonSpec,
 	masterConnectionSpec ytv1.MasterConnectionSpec,
+	masterCachesSpec ytv1.MasterCachesSpec,
```
l0kix2, line 25: It is inconsistent that one `New*` requires pointer and other is not.

pkg/ytconfig/generator.go
```diff
@@ -127,9 +127,8 @@ func (g *Generator) fillDriver(c *Driver) {
 
 	c.PrimaryMaster.Addresses = g.getMasterAddresses()
 	c.PrimaryMaster.CellID = generateCellID(g.ytsaurus.Spec.PrimaryMasters.CellTag)
-
-	c.MasterCache.EnableMasterCacheDiscover = true
-	g.fillPrimaryMaster(&c.MasterCache.MasterCell)
+	//c.PrimaryMaster.Peers = g.getMasterHydraPeers()
```
l0kix2, line 130: remove commented?

pkg/ytconfig/generator.go
```diff
@@ -693,3 +698,53 @@ func (g *Generator) GetDiscoveryConfig() ([]byte, error) {
...
 }
+
+func (g *Generator) getMasterCachesConfigImpl() (MasterCacheServer, error) {
+	spec := g.ytsaurus.Spec.MasterCaches
+	c, err := getMasterCachesCarcass(spec)
+	if err != nil {
+		return MasterCacheServer{}, err
+	}
+	g.fillCommonService(&c.CommonServer, &spec.InstanceSpec)
+	//c.ClusterConnection.MasterCache.Addresses = g.getMasterCachesAddresses()
```
l0kix2, line 709: remove commented?

pkg/ytconfig/generator_test.go
```diff
@@ -575,3 +608,22 @@ func getMasterConnectionSpecWithFixedMasterHosts() v1.MasterConnectionSpec {
 	spec.CellTag = 1000
 	return spec
 }
+
+//func getMasterCachesConnectionSpec() v1.MasterCachesConnectionSpec {
+//	return v1.MasterCachesConnectionSpec{}
+//}
```
l0kix2, line 614: remove?

pkg/ytconfig/generator.go
```diff
@@ -155,6 +154,12 @@ func (g *BaseGenerator) fillClusterConnection(c *ClusterConnection, s *ytv1.RPCT
 	c.ClusterName = g.key.Name
 	c.DiscoveryConnection.Addresses = g.getDiscoveryAddresses()
 	g.fillClusterConnectionEncryption(c, s)
+	if len(g.getMasterCachesAddresses()) == 0 {
+		c.MasterCache.Addresses = g.getMasterAddresses()
+	} else {
+		c.MasterCache.Addresses = g.getMasterCachesAddresses()
+	}
```
l0kix2, line 161: Looks like default behaviour to go to master if master caches is empty. @gritukan is it better somehow to fill config with master addresses for caches for some reason?

pkg/components/master_caches.go
qurname2: Done

pkg/ytconfig/node_generator.go
```diff
@@ -15,12 +15,15 @@ func NewRemoteNodeGenerator(
 	clusterDomain string,
 	commonSpec ytv1.CommonSpec,
 	masterConnectionSpec ytv1.MasterConnectionSpec,
+	masterCachesSpec ytv1.MasterCachesSpec,
+	// masterCachesConnectionSpec ytv1.MasterCachesConnectionSpec,
```
l0kix2, line 19: remove?

api/v1/test_helpers.go
qurname2: Fixed

l0kix2 (changes requested):
I think there are only minor issues left.

Makefile
qurname2: It's useful for local testing. Let's say you don't want to recreate kind cluster, but would like to upgrade operator's image. In that case you have to specify a new OPERATOR_TAG in Makefile and run `make helm-kind-install`. Operator's image will be upgraded, using helm

l0kix2: You mean to use some prebuilt operator image, not the one we building from source?

l0kix2: Maybe better to get OPERATOR_TAG from envvar with current default so you don't need to edit Makefile for that action

api/v1/test_helpers.go
qurname2: I've changed the order of the images

pkg/ytconfig/base_generator.go
qurname2: Fxied

api/v1/test_helpers.go
l0kix2: looks like unresolved

qurname2: Sorry, forgot to remove it

l0kix2 (approved):

PR #123, opened 2024-02-23
Do not serialize zero values for quota parameters

psushin:
When quota parameters for locations could not be discovered from the PVC requests, they default to zero, which ytsaurus nodes perceive as a real zero-quota. Instead we prefer not to serialize these config parameters, and rely on the default behavior.


l0kix2 (approved):
LGTM

PR #140, opened 2024-03-01
Add yqla token file in container

Krisha11:

### DIFF
--- a/pkg/components/yql_agent.go
+++ b/pkg/components/yql_agent.go
@@ -143,7 +143,7 @@ func (yqla *yqlAgent) doSync(ctx context.Context, dry bool) (ComponentStatus, er
 		if !dry {
 			ss := yqla.server.buildStatefulSet()
 			container := &ss.Spec.Template.Spec.Containers[0]
-			container.Command = []string{"sh", "-c", fmt.Sprintf("echo $YT_TOKEN > %s; %s", consts.DefaultYqlTokenPath, strings.Join(container.Command, " "))}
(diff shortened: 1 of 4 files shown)


pkg/components/yql_agent.go
```diff
@@ -141,8 +141,10 @@ func (yqla *yqlAgent) doSync(ctx context.Context, dry bool) (ComponentStatus, er
 
 	if yqla.NeedSync() {
 		if !dry {
+			token, _ := yqla.secret.GetValue(consts.TokenSecretKey)
 			ss := yqla.server.buildStatefulSet()
 			container := &ss.Spec.Template.Spec.Containers[0]
+			container.Command = []string{"sh", "-c", fmt.Sprintf("echo -n %s > /usr/yql_agent_token; %s", token, strings.Join(container.Command, " "))}
```
gritukan, line 147: Won't token be shown in `ps` or pod description in this case?

I suggest to use environment variable. Looks like it is already passed, so `echo $SECRET >/usr/yql_agent_token` should be sufficient.

gritukan, line 147: Shouldn't we use `YTTokenPath` instead of hardcoded `/usr/yql_agent_token` here?

Krisha11: It is not very easy to access the config here. It has already been serialized by this point. So I put token path in consts

l0kix2:
@gritukan do you have any notes here or we can merge it?

gritukan (approved):

PR #141, opened 2024-03-04
regenerate crd

koct9i:


l0kix2 (approved):

l0kix2:
My bad, there didn't run yaml generating in my PR after @kruftik PR was merged.
We need to add a check for if there is generated files diff.

PR #147, opened 2024-03-05
configurable monitoringPort

### LINKED ISSUE #146: fix k8s service and readinessProbe specs in case of monitoring port is overridden via configmap

kruftik:
fixes #146


l0kix2 (changes requested):
As far as I can see default monitoring port is used in the config [generators](https://github.com/ytsaurus/yt-k8s-operator/blob/ccc3c60e9d02ffb6b8912f8415e3c43883756355/pkg/ytconfig/master.go#L53-L53). I suppose it also should be taken from InstanceSpec if specified, isn't it?

If so, the default port must be set in spec in the very beginning of reconciliation loop. 
Maybe the best way is to implement the default settings webhook 
https://book.kubebuilder.io/cronjob-tutorial/webhook-implementation
? If this works we can explicitly set ports to defaults in the spec at the very start and not worrying where to override them later.

l0kix2:
And what about usage of the monitoring port in ytconfig i've mentioned earlier. Shouldn't it be configured consistently?

pkg/components/data_node.go
```diff
@@ -29,7 +30,10 @@ func NewDataNode(
 		APIProxy:       ytsaurus.APIProxy(),
 		ComponentLabel: cfgen.FormatComponentStringWithDefault(consts.YTComponentLabelDataNode, spec.Name),
 		ComponentName:  cfgen.FormatComponentStringWithDefault("DataNode", spec.Name),
-		MonitoringPort: consts.DataNodeMonitoringPort,
```
koct9i, line 32: You could leave monitoring port on labeler but fetch it from spec:

`MonitoringPort: ptr.Int32Deref(spec.InstanceSpec.MonitoringPort, consts.DataNodeMonitoringPort)`

pkg/components/discovery.go
```diff
@@ -23,7 +24,10 @@ func NewDiscovery(cfgen *ytconfig.Generator, ytsaurus *apiproxy.Ytsaurus) *Disco
 		APIProxy:       ytsaurus.APIProxy(),
 		ComponentLabel: consts.YTComponentLabelDiscovery,
 		ComponentName:  "Discovery",
-		MonitoringPort: consts.DiscoveryMonitoringPort,
+	}
+
+	if resource.Spec.Discovery.InstanceSpec.MonitoringPort == nil {
+		resource.Spec.Discovery.InstanceSpec.MonitoringPort = ptr.Int32(consts.DiscoveryMonitoringPort)
```
koct9i, line 30: IMHO changing spec isn't a good idea.

pkg/components/data_node.go
kruftik: I intentionally dropped it from here since I don't see any reason to be inside. Monitoring/api port seems to have neither relation to the 'labeller' (how the struct is named) logic

koct9i: Yeah, maybe it needs better name or refactoring,
but this pr cloud be much smaller with leaving port here.

pkg/components/discovery.go
l0kix2: I've suggested to make make a function, which would update all instanceSpecs in ytsaurusSpec, for start it can
  - sets monitoring port to default if not specified
  - sets image to coreImage if not specified
This function should be called early in the reconcillation loop.
That way all our code can expect actual values in InstanceSpec and get rid of ifs.

For a cleaner solution here may be introduced some new struct, in which we will put values from spec/instanceSpec with beforementined logic and use this struct, not the original instanceSpec later. But i suppose this would be rather big change, not sure if it worth it.

kruftik: well, external data mutation by pointer is not the best approach for sure. what you can propose here? would it be better to accept a struct rather than a pointer?

api/v1/ytsaurus_types.go
```diff
@@ -228,7 +228,7 @@ type InstanceSpec struct {
 	Volumes      []corev1.Volume      `json:"volumes,omitempty"`
 	VolumeMounts []corev1.VolumeMount `json:"volumeMounts,omitempty"`
 	//+optional
-	ReadinessProbeParams  HealthcheckProbeParams          `json:"readinessProbeParams,omitempty"`
+	ReadinessProbeParams  *HealthcheckProbeParams         `json:"readinessProbeParams,omitempty"`
```
l0kix2, line 231: Next time please do separate PR for independent issues, that could be merged ages ago.

l0kix2 (approved):
I'm ok this this change for now, though, as we've discussed, it is undesirable to change instanceSpec in the middle of code. But improving that may require some refactoring, hope will do that later.

PR #168, opened 2024-03-18
added WATCH_NAMESPACE env and LeaderElectionNamespace

qurname2:
added WATCH_NAMESPACE env to the helm chart and LeaderElectionNamespace option when WATCH_NAMESPACE is exist

### DIFF
--- a/main.go
+++ b/main.go
@@ -106,6 +106,7 @@ func main() {
 			}
 		} else {
 			managerOptions.Cache.DefaultNamespaces[watchNamespace] = cache.Config{}
+			managerOptions.LeaderElectionNamespace = watchNamespace
 		}
 	}
 
--- a/main.go
+++ b/main.go
@@ -18,10 +18,9 @@ package main
 
 import (
 	"flag"
+	"go.uber.org/zap/zapcore"
 	"os"
 	"strings"
-
-	"go.uber.org/zap/zapcore"
 
 	"github.com/ytsaurus/yt-k8s-operator/controllers"
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
@@ -98,7 +97,9 @@ func main() {
 	}
 
 	watchNamespace, ok := os.LookupEnv("WATCH_NAMESPACE")
-	if ok {
+	// We can't setup managerOptions.Cache.DefaultNamespaces = map[cache.AllNamespaces]cache.Config{} due to
+	// https://github.com/kubernetes-sigs/controller-runtime/issues/2628
+	if ok && watchNamespace != "nil" {
 		managerOptions.Cache.DefaultNamespaces = map[string]cache.Config{}
 		if strings.Contains(watchNamespace, ",") {
 			for _, namespace := range strings.Split(watchNamespace, ",") {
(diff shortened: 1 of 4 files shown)


l0kix2:
@qurname2 let's rebase and resolve conflicts
@koct9i cloud you take a look at couple of lines.

ytop-chart/templates/deployment.yaml
```diff
@@ -47,6 +47,10 @@ spec:
           value: {{ quote .Values.controllerManager.manager.env.ytLogLevel }}
         - name: KUBERNETES_CLUSTER_DOMAIN
           value: {{ quote .Values.kubernetesClusterDomain }}
+        {{- if .Values.controllerManager.manager.watchNamespace }}
+        - name: WATCH_NAMESPACE
+          value: {{ .Values.controllerManager.manager.watchNamespace }}
+        {{- end }}
```
koct9i, line 53: This file is generated.

Possible environment variables are defined in config/manager/manager.yaml

In trunk watched namespaces are controlled by map DefaultNamaspaces
https://github.com/kubernetes-sigs/controller-runtime/blob/v0.16.5/pkg/cache/cache.go#L188

Where "" is = AllNamespaces
https://github.com/kubernetes-sigs/controller-runtime/blob/v0.16.5/pkg/cache/cache.go#L128

l0kix2:
Can someone remind me if we wanted to close that and do namespesify in other places or it should be finished?

main.go
```diff
@@ -98,14 +97,17 @@ func main() {
 	}
 
 	watchNamespace, ok := os.LookupEnv("WATCH_NAMESPACE")
-	if ok {
+	// We can't setup managerOptions.Cache.DefaultNamespaces = map[cache.AllNamespaces]cache.Config{} due to
+	// https://github.com/kubernetes-sigs/controller-runtime/issues/2628
+	if ok && watchNamespace != "nil" {
```
koct9i, line 102: Magic string value "nil" is _really_ bad choice, it looks like null but doesn't.
Empty string should work fine here.

Constant "AllNamespaces" is exactly empty string because it is invalid namespace name and natural "null" value for strings. While "nil" is odd, but still valid name.

main.go
```diff
@@ -98,14 +97,17 @@ func main() {
...
 		managerOptions.Cache.DefaultNamespaces = map[string]cache.Config{}
 		if strings.Contains(watchNamespace, ",") {
 			for _, namespace := range strings.Split(watchNamespace, ",") {
 				managerOptions.Cache.DefaultNamespaces[namespace] = cache.Config{}
 			}
 		} else {
 			managerOptions.Cache.DefaultNamespaces[watchNamespace] = cache.Config{}
+			managerOptions.LeaderElectionNamespace = watchNamespace
```
koct9i, line 110: Ok, this allows to isolate leader-election and watched/managed namespace.

This are still may be conflict in webhooks.
I guess, kube-api will call validating/mutating hooks for all installed operators.

We could solve that in separate PR.

koct9i (changes requested):

main.go
qurname2: Okey, will change it to ""

koct9i (approved):

PR #169, opened 2024-03-18
added discovery_server section into master config

qurname2:
to resolve https://github.com/ytsaurus/yt-k8s-operator/issues/63

### DIFF
--- a/pkg/ytconfig/master.go
+++ b/pkg/ytconfig/master.go
@@ -28,6 +28,10 @@ type CypressManager struct {
 	DefaultJournalWriteQuorum       int `yson:"default_journal_write_quorum,omitempty"`
 }
 
+type DiscoveryServers struct {
+	AddressList
+}
+
 type MasterServer struct {
 	CommonServer
 	Snapshots        MasterSnapshots  `yson:"snapshots"`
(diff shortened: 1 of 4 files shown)


l0kix2 (approved):

pkg/ytconfig/master.go
```diff
@@ -37,6 +41,7 @@ type MasterServer struct {
 	CypressManager   CypressManager   `yson:"cypress_manager"`
 	PrimaryMaster    MasterCell       `yson:"primary_master"`
 	SecondaryMasters []MasterCell     `yson:"secondary_masters"`
+	DiscoveryServers DiscoveryServers `yson:"discovery_server"`
```
achulkov2, line 44: Let's keep the naming consistent with the yson config name. So DiscoveryServer without the plural.

achulkov2:
Let's add a check to one of our e2e tests that checks that the directory `//sys/discovery_servers` is filled with orchids.

achulkov2 (changes requested):

l0kix2:
@qurname2 it looks like just small fixes left here. Maybe you will finish it some time (obviously not urgent if it already hangs for too long) or close if already not needed?

PR #177, opened 2024-03-21
Remove ENV YT_AUTH_CLUSTER_ID

kozubaeff:


l0kix2:
Shouldn't we do this in different releases of the operator with mentioning what UI version is supported in each?

PR #182, opened 2024-03-23
labeller: make component labels more descrete

kruftik:


pkg/components/chyt.go
```diff
@@ -34,13 +34,14 @@ func NewChyt(
...
-		Annotations:    ytsaurus.Spec.ExtraPodAnnotations,
-	}
+	l := labeller.NewLabeller(
+		&chyt.GetResource().ObjectMeta,
+		consts.CHYTType,
+		consts.YTComponentLabelCHYT,
+	).
+		WithComponentInstanceName(chyt.GetResource().Name).
+		WithAnnotations(ytsaurus.Spec.ExtraPodAnnotations).
+		Build()
```
koct9i, line 44: Half constructor, half builder. Looks really ugly.

koct9i (changes requested):

pkg/components/chyt.go
l0kix2: As far as I understood we've agreed to have two (or more) constructors to cover of possible cases of creation with fields required for that case..

pkg/labeller/labeller.go
```diff
@@ -16,39 +15,79 @@ type FetchableObject struct {
...
+
+		annotations: annotations,
+	}
+
+	if instanceName != "" && instanceName != consts.DefaultName {
+		l.ComponentFullName = fmt.Sprintf("%s-%s", nameBase, instanceName)
+		l.ComponentObjectsNamePrefix = fmt.Sprintf("%s-%s", objectsNamePrefixBase, instanceName)
+	}
+
+	return l
```
l0kix2, line 62: Not sure if some style guide mentions it, but I think it is good practice to write constructors like that:
```go
func NewLabeller(...) Labeller {
    // variables and ifs
    return Labeller{
       ...
    }
}
```
than creating an instance, and redefining its fields later.

pkg/labeller/labeller.go
```diff
@@ -16,39 +15,79 @@ type FetchableObject struct {
...
-	ComponentName  string
-	Annotations    map[string]string
+	ObjectMeta                 *metav1.ObjectMeta
+	ComponentObjectsNamePrefix string
+	ComponentFullName          string
+	annotations                map[string]string
+
+	instanceName string
+
+	objectsNamePrefixBase string
```
l0kix2, line 25: Let's add description for this fields here with examples so it would be obvious to understand what values to expect in those fields for singleton and multiinstance components.

For now it is hard to say what will be in values and no tests to see what they are.

pkg/components/strawberry_controller.go
```diff
@@ -66,13 +66,12 @@ func NewStrawberryController(
...
-	l := labeller.Labeller{
-		ObjectMeta:     &resource.ObjectMeta,
-		APIProxy:       ytsaurus.APIProxy(),
-		ComponentLabel: fmt.Sprintf("yt-%s-controller", name),
-		ComponentName:  componentName,
-		Annotations:    resource.Spec.ExtraPodAnnotations,
-	}
+	l := labeller.NewLabellerForGlobalComponent(
+		&resource.ObjectMeta,
+		consts.ComponentType(componentName),
```
l0kix2, line 71: Shouldn't CHYTType be here?
The point of `ComponentType` is that it is a finite number of well-known values and not expected to created them dynamically.

pkg/consts/labels.go
```diff
@@ -18,4 +18,8 @@ const (
 	YTComponentLabelYqlAgent        string = "yt-yql-agent"
 	YTComponentLabelClient          string = "yt-client"
 	YTComponentLabelMasterCache     string = "yt-master-cache"
+	YTComponentLabelQueryTracker    string = "yt-query-tracker"
+	YTComponentLabelQueueAgent      string = "yt-queue-agent"
+	YTComponentLabelSPYT            string = "ytsaurus-spyt"
+	YTComponentLabelCHYT            string = "ytsaurus-chyt"
```
l0kix2, line 24: Will something breaks (k8s labels?) if these are will be called `yt-spyt` and `yt-chyt` also?
Do we want to have them consistent with others or it is intended?

l0kix2 (changes requested):
Since those names of resources and labels are kinda part of our public API (some users might write scripts expecting them) and if we don't want to break it now or in the future it would be wise to first make a PR with test for this values. I think ComponentManager constructor should be split in two parts, where first one is about creating components and not interacting with k8s — so that constructor can be used in such test and we call `GetMetaLabelMap` for all components and canonize the result. 
After that this PR will introduce change and we can be sure if nothing is broken and new changes are intended.

pkg/consts/labels.go
kruftik: well, I didn't dive into deep enough, it's just string constants moved out from component constructors to be aligned with the rest of component types

pkg/components/strawberry_controller.go
kruftik: good point, missed it

l0kix2:
I guess this one also not actual now after labels refactoring.

PR #184, opened 2024-03-25
WIP: support for extra labels configuring via component spec

kruftik:
worth looking at after #182


koct9i:
Is there any use-case for per-component labels?

PR #185, opened 2024-03-25
Tune tablet changelog/snapshot initial replication factor according to data node count

koct9i:
Use 1 replica for less than 3 data nodes, and default 3 replicas otherwise.

Link: https://github.com/ytsaurus/ytsaurus/blob/b9b7e0918d2141a5c041b217bcd6cc1b7ff7e8d9/yt/yt/tests/library/yt_env_setup.py#L1684


koct9i:
Table bundle "default" is still created with replication factor 3, but this doesn't block cluster bootstrap.

PR #189, opened 2024-03-26
Parallelize e2e tests

koct9i:
- e2e: use ginkgo context and defer
  
- e2e: use node ipv4 address
  
- e2e: run test by ginkgo

- e2e: run tests in parallel

### DIFF
--- a/test/e2e/suite_test.go
+++ b/test/e2e/suite_test.go
@@ -28,8 +28,9 @@ import (
 	. "github.com/onsi/gomega"
 
 	"k8s.io/client-go/kubernetes/scheme"
-	"k8s.io/client-go/rest"
+
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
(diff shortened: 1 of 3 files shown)


koct9i:
Right now speed up is 32m -> 22m - e2e part 15m -> 5m

test/e2e/suite_test.go
```diff
@@ -41,11 +42,8 @@ import (
 // These tests use Ginkgo (BDD-style Go testing framework). Refer to
 // http://onsi.github.io/ginkgo/ to learn more about Ginkgo.
 
-var cfg *rest.Config
 var k8sClient client.Client
-var testEnv *envtest.Environment
 var ctx context.Context
```
l0kix2, line 46: Can we get rid of this package level ctx and use ctx which is passed in suite by ginkgo?

l0kix2 (approved):
Tests speed up is much appreciated!

test/e2e/suite_test.go
koct9i: Yep, will do later.

Issue #181, opened 2024-03-22
QT ACO namespace configuration


Krock21:
We need to set good default settings for ACO namespace `queries` that will be created during init job

Suggested format:
```
$ yt get //sys/access_control_object_namespaces/queries/@acl
[
    {
        "action" = "allow";
        "subjects" = [
            "owner";
        ];
        "permissions" = [
            "read";
            "write";
            "administer";
            "remove";
        ];
        "inheritance_mode" = "immediate_descendants_only";
    };
    {
        "action" = "allow";
        "subjects" = [
            "users";
        ];
        "permissions" = [
            "modify_children";
        ];
        "inheritance_mode" = "object_only";
    };
]

$ yt get //sys/access_control_object_namespaces/queries/nobody/@principal_acl
[]

$ yt get //sys/access_control_object_namespaces/queries/everyone/@principal_acl
[
    {
        "action" = "allow";
        "subjects" = [
            "everyone";
        ];
        "permissions" = [
            "read";
            "use";
        ];
        "inheritance_mode" = "object_and_descendants";
    };
]

$ yt get //sys/access_control_object_namespaces/queries/everyone-use/@principal_acl
[
    {
        "action" = "allow";
        "subjects" = [
            "everyone";
        ];
        "permissions" = [
            "use";
        ];
        "inheritance_mode" = "object_and_descendants";
    };
]
```

l0kix2:
@Krock21 should this issue be closed after #195?

l0kix2:
Resolved in [0.7.0](https://github.com/ytsaurus/yt-k8s-operator/releases/tag/release%2F0.7.0)

PR #195, opened 2024-03-29
Init query tracker state on cluster update, Fix access control for query tracker ACOs

Krock21:
Issue: #181 

Fixing "queries", "everyone" ACLs. Adding everyone-use ACO
Creating them even on cluster update, if they didn't exist before

We don't want to override manual changes made by cluster admins on update, hence it does nothing if objects exist

### DIFF
--- a/pkg/components/query_tracker.go
+++ b/pkg/components/query_tracker.go
@@ -169,18 +169,15 @@ func (qt *QueryTracker) doSync(ctx context.Context, dry bool) (ComponentStatus, 
 	}
 
 	var ytClient yt.Client
+	if !dry {
+		ytClient = qt.ytsaurusClient.GetYtClient()
+		if ytClient == nil {
+			return WaitingStatus(SyncStatusPending, "getting yt client"), err
+		}
+	}
+
 	if qt.ytsaurus.GetClusterState() != ytv1.ClusterStateUpdating {
-		ytClientStatus, err := qt.ytsaurusClient.Status(ctx)
-		if err != nil {
-			return ytClientStatus, err
-		}
-		if ytClientStatus.SyncStatus != SyncStatusReady {
-			return WaitingStatus(SyncStatusBlocked, qt.ytsaurusClient.GetName()), err
-		}
-
 		if !dry {
-			ytClient = qt.ytsaurusClient.GetYtClient()
-
 			err = qt.createUser(ctx, ytClient)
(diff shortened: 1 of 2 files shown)


pkg/components/query_tracker.go
```diff
@@ -304,6 +302,20 @@ func (qt *QueryTracker) init(ctx context.Context, ytClient yt.Client) (err error
 		&yt.CreateObjectOptions{
 			Attributes: map[string]interface{}{
 				"name": "queries",
+				"acl": []interface{}{
+					map[string]interface{}{
+						"action":           "allow",
+						"subjects":         []string{"owner"},
```
Krisha11, line 308: Operator does not create groups and users with "owner" name. Maybe it will be better to give the rights to the "admin" user

https://github.com/ytsaurus/yt-k8s-operator/blob/main/pkg/consts/defaults.go#L3

pkg/components/query_tracker.go
```diff
@@ -349,6 +361,27 @@ func (qt *QueryTracker) init(ctx context.Context, ytClient yt.Client) (err error
 		logger.Error(err, "Creating access control object 'everyone' in namespace 'queries' failed")
 		return
 	}
+
+	_, err = ytClient.CreateObject(
+		ctx,
+		yt.NodeAccessControlObject,
+		&yt.CreateObjectOptions{
+			Attributes: map[string]interface{}{
+				"name":      "everyone-use",
```
Krisha11, line 370: It would be great to add new ACOs to the documentation later

pkg/components/query_tracker.go
Krisha11: Read that this is a special fictitious user

Krock21: https://ytsaurus.tech/docs/ru/user-guide/storage/access-control#owner

`owner` is a built-in user: https://github.com/ytsaurus/ytsaurus/blob/4979a8c88ac626cdd3e973618c5f1a4d65cca577/yt/yt/server/master/security_server/security_manager.cpp#L3825-L3826

`admins` get access from inherited `access_control_object_namespace` ACL
`owner` will be substituted by creator of a new ACO and will grant access to creator

Krock21:
Fixed behaviour on update

Krisha11 (approved):

l0kix2 (approved):

PR #197, opened 2024-04-03
Enable rpc proxy in job proxy

Alexvsalexvsalex:


pkg/ytconfig/node.go
```diff
@@ -492,10 +494,14 @@ func getExecNodeServerCarcass(spec *ytv1.ExecNodesSpec, commonSpec *ytv1.CommonS
 	jobProxyLoggingBuilder.logging.FlushPeriod = 3000
 	c.ExecNode.JobProxy.JobProxyLogging = jobProxyLoggingBuilder.logging
 
+	c.ExecNode.JobProxy.JobProxyAuthenticationManager.RequireAuthentication = true
```
l0kix2, line 497: Is it backward compatible behaviour or users will need to configure that in their existing specs so their ytsaurus would start after the operator update?

Alexvsalexvsalex: The PR set up correct configuration of RPC proxy inside of Job proxy (new opt-in feature that has no active users beside SPYT). I found that the feature doesn't work at all without these changes. So it just provides the fix for a bug.

l0kix2: So to be clear, `job_proxy_authentication_manager/require_authentication` doesn't require extra configuration for exec nodes to work?
And currently using rpc proxy inside jobs is not working, correct?

koct9i (approved):

pkg/ytconfig/node.go
Alexvsalexvsalex: 1) Well, as you can see I also enabled `cypress_token_authenticator`, so there is authentication by cypress token which is default way. In my opinion it's enough to have now. Default other configs are applicable here.
2) Yeah, currently using the feature will just lead errors and operation fail. It has been found by user's issue.

PR #203, opened 2024-04-08
Enable some linters and fix issues

koct9i:
- Update golangci-lint config
  
- Cleanup api/v1/ytsaurus_webhook.go
  
- Fix lint issues
  
- Fix lint issues

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,2 +1,244 @@
 run:
   timeout: 5m
+  allow-parallel-runners: true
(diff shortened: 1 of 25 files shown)


l0kix2 (approved):
LGTM, though it would be better to rebase on main and have green tests before merge

l0kix2:
Tests are failing, maybe timeout for update should be increased? jobs/tablet cells deletions work rather long at my laptop usually.

PR #217, opened 2024-04-10
Fix running jobs with custom docker_image when default job image is not set

koct9i:
Logic behind UseJobProxyFromImage is slightly broken. Keep it disabled for now.


l0kix2 (approved):
but test should be recanonized before merge

PR #223, opened 2024-04-12
Use info stderr logger in tests

koct9i:


l0kix2 (approved):
why don't we have stdout btw?

PR #225, opened 2024-04-12
Add option for creating ephemeral cluster

koct9i:
- Add option for creating ephemeral cluster
  This is high-level option which could effects on its own or
  allow other "dangerous" options.
  
  For now it simply disables "sync" in all disk I/O engines.
  
- Use ephemeral clusters in tests


l0kix2 (approved):
Hope the tests will speed up

Issue #115, opened 2024-02-06
Refactor operator reconcilliation flow


l0kix2:
Currently we have an implementation where ytsaurus cluster update flow is hard to understand because it is rather complex and scattered across codebase. As a consequence it is error-prone, hard to maintain and operate on real clusters.

We want to simplify things by implementing an approach where components will be brought to the desired state one after another in strict linear sequence. I.e each reconciliation loop will go through that (well known and explicitly described in a single place of the codebase) sequence, find first component which is NOT in desired state and execute one action which should bring that component closer to the desired state.

We expect that such flow would be easier to understand and also it would be easy to troubleshoot in what stage of update operator and cluster currently are.

As a downside of the approach we expect that linear update of components would be slower, which is might be thing to improve in the future.

l0kix2:
After today's discussion https://github.com/ytsaurus/ytsaurus-k8s-operator/issues/396 it seems that we still need 2 update flows: full (as it works now with all pre-steps, after that all pods are removed and after that all after steps) and linear (where only pods of one component are removed at each moment of flow).
I guess we can introduce some flow steps and build 2 different update flows (and create flow i guess) using them.

PR #240, opened 2024-04-16
UpdateSelector: add field

l0kix2:
Introduced new `UpdateSelector` field in spec which can be one of: 
    - `""` (empty) — field is not considered at all, which is backward compatible with using `EnableFullUpdate` 
    - `Nothing` — no component can be updated (logs will indicate a reason)
    - `MasterOnly` — only master will be updated (flow with safety checks/safe mode/snapshots/exit read only), other components' updates will be ignored
    - `TabletNodesOnly` — only tablet nodes will be updated (with removing tablet bundles and recreating them after the update)
    - `ExecNodesOnly` — only exec nodes will be updated (without extra steps, because it is stateless component)
    - `StatelessOnly` — everything apart from master and tablet nodes will be updated (currently it is the same as so called local update flow)
    - `Everything ` — everything will be updated at once (currently it is the same as so called Full update flow)
I've called the new field experimental since I'm not sure if we will want to support all that behaviours in the future after #115 

Also `Flow` field is added to `UpdateStatus` which is persisted with transition to the Updating status. Usage can be seen in #239

### DIFF
--- a/api/v1/ytsaurus_types.go
+++ b/api/v1/ytsaurus_types.go
@@ -547,9 +547,9 @@ type YtsaurusSpec struct {
 	//+optional
 	EnableFullUpdate bool `json:"enableFullUpdate"`
 	//+optional
-	// UpdateStrategy is an experimental field. Behaviour may change.
-	// If UpdateStrategy is not empty EnableFullUpdate is ignored.
-	UpdateStrategy UpdateStrategy `json:"updateStrategy"`
+	// UpdateSelector is an experimental field. Behaviour may change.
+	// If UpdateSelector is not empty EnableFullUpdate is ignored.
+	UpdateSelector UpdateSelector `json:"updateSelector"`
 
 	Bootstrap *BootstrapSpec `json:"bootstrap,omitempty"`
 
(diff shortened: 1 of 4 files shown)


api/v1/ytsaurus_types.go
```diff
@@ -618,10 +622,33 @@ type TabletCellBundleInfo struct {
...
+	UpdateFlowMaster      UpdateFlow = "Master"
+	UpdateFlowTabletNodes UpdateFlow = "TabletNodes"
+	UpdateFlowFull        UpdateFlow = "Full"
+)
+
 type UpdateStatus struct {
 	//+kubebuilder:default:=None
 	State                 UpdateState            `json:"state,omitempty"`
 	Components            []string               `json:"components,omitempty"`
+	Flow                  UpdateFlow             `json:"flow,omitempty"`
```
koct9i, line 651: Please add comment that this field means - it goes into CRD and api.md

api/v1/ytsaurus_types.go
```diff
@@ -618,10 +622,33 @@ type TabletCellBundleInfo struct {
 	TabletCellCount int    `yson:"tablet_cell_count,attr" json:"tabletCellCount"`
 }
 
+type UpdateSelector string
+
+const (
+	UpdateSelectorNone            UpdateSelector = ""
```
koct9i, line 628: "None" is confusing, especially when it is not "Nothing".
Maybe "Unspecified".

koct9i (approved):

api/v1/ytsaurus_types.go
l0kix2: done

api/v1/ytsaurus_types.go
l0kix2: done

PR #245, opened 2024-04-16
UpdateSelector: tests

### LINKED ISSUE #211: Implement master update without full update
Currently master image update requires all other components' pods to be recreated also. It is not always needed, so lets support master update strategy without full update (only master pods should be recreated, but as in full update snapshots should be built).

l0kix2:
Fixes #211

Added the tests for each of the used selectors with checking which pods exactly was updated (and which are not).

### DIFF
--- a/test/e2e/ytsaurus_controller_test.go
+++ b/test/e2e/ytsaurus_controller_test.go
@@ -263,8 +263,7 @@ var _ = Describe("Basic test for Ytsaurus controller", func() {
 			getSimpleUpdateScenario("test-major-update", ytv1.CoreImageNextVer),
 		)
 		It(
-			"Should be updated according to UpdateSelector=Everything",
-			func(ctx context.Context) {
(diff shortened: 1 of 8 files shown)


test/e2e/ytsaurus_controller_test.go
```diff
@@ -261,7 +260,156 @@ var _ = Describe("Basic test for Ytsaurus controller", func() {
 			"Should run and update Ytsaurus to the next major version",
 			getSimpleUpdateScenario("test-major-update", ytv1.CoreImageNextVer),
 		)
+		It(
+			"Should be updated according to UpdateSelector=Everything",
+			func(ctx context.Context) {
```
koct9i, line 265: Weird indentation. You can put It("description", func(...) {" into one line.

If linter complains about long line we can tweak it for tests.

koct9i (approved):

PR #257, opened 2024-05-02
Add support for directDownload to UI Spec

kozubaeff:
closes  https://github.com/ytsaurus/yt-k8s-operator/issues/256


api/v1/ytsaurus_types.go
```diff
@@ -452,6 +452,8 @@ type UISpec struct {
 	Theme       string  `json:"theme,omitempty"`
 	Description *string `json:"description,omitempty"`
 	Group       *string `json:"group,omitempty"`
+
+	DirectDownload *bool `json:"directDownload,omitempty"`
```
koct9i, line 456: Please add comment above that this option does - it goes into CRD and API doc.

What kind of "direct" is this?
Operations goes via YT http proxy not via UI backend?

kozubaeff: added a comment and recreated the docs

l0kix2: I think the preferable way for operator docs is to have a link to UI code/docs with description what the option does and just mention that operator simply puts field in ui config.

Though if sending PR to UI code seem complicated let's at least describe all options if we choose to use bool pointer — we have to document all three values: false, true, omitted. 
I suppose docs should look like `If omitted or set to true X happend and if set to false — Y happens`.
Also it is not clear what UI is downloading.

l0kix2 (changes requested):

api/v1/ytsaurus_types.go
```diff
@@ -452,6 +452,8 @@ type UISpec struct {
 	Theme       string  `json:"theme,omitempty"`
 	Description *string `json:"description,omitempty"`
 	Group       *string `json:"group,omitempty"`
+	// When this is set to false, UI will use backend for downloading instead of proxy.
```
l0kix2, line 455: As I mentioned here https://github.com/ytsaurus/yt-k8s-operator/pull/257#discussion_r1604919789 this comment should help users to choose the right value. Since it is a bool pointer we have three possible values here: nil, true, and false.
You have only documented only the `false` option, so when users update their operator they may have reasonable question: currently my spec doesn't have such field, so what happens when I update the operator with the same spec. It is not obvious to me what "direct" means here, so I suppose it may not be obvious for users (suppose if false means backend then proxy is the direct one). Usually `nil` and `false` values mean the same, so as a user, I would suggest that if I don't set the value I get UI with changed behaviour. But at the same time for some reason you documenting the only `false` option, so maybe (and I think that is the case) the default is `true`.

If my guess is right and this (poorly named in UI) option is true for default — there is more reason to document it clearly. So if true is really a default for UI config let us say something like:
```
When DirectDownload is true or omitted, UI uses the proxies for downloading, which is the default behaviour.
If set to false, UI uses its backend for downloading.
```

It all would be less confusing if settings was named `BackendDownload`, but since it is in UI public configuration let's stick with `DirectDownload` and not add entropy.

l0kix2 (changes requested):
tl;dr:
1) if the new option is not set — we should have zero diff in the UI configs
2) it should be possible for users to understand what happens when each of three possible values are chosen for the field.

kozubaeff:
Thought using name DisableDirectDownload would be more convenient. Recreated the docs and generated the test which checks for backward compatibility

api/v1/ytsaurus_types.go
```diff
@@ -457,6 +457,9 @@ type UISpec struct {
 	Theme       string  `json:"theme,omitempty"`
 	Description *string `json:"description,omitempty"`
 	Group       *string `json:"group,omitempty"`
+	// When this is set to true, UI will use backend for downloading instead of proxy. If this is set to false or omitted, UI use proxies, which is a default behaviour.
+	//+optional
+	DisableDirectDownload bool `json:"disableDirectDownload,omitempty"`
```
l0kix2, line 462: Usually, in golang code I'm in favor of not using pointers if nil doesn't have special meaning for us. 
But our practice shows that in operator crd specs it is better to use pointers. I suggest we better revert to `*bool`, because if we use non-pointer `bool` `kubectl apply` will show diff `disableDirectDownload: false` for users who haven't changed anything in spec and don't expect diff. Writing down `false` default value in the applied spec is also not a very good idea, because it makes our future possible default change decisions problematic.

l0kix2 (changes requested):

l0kix2 (approved):

PR #258, opened 2024-05-02
Add configuration for solomon exporter: specify host and some instance tags

achulkov2:
When the solomon_exporter section is not configured metrics are exported without any host-related labels, which isn't very convenient.

Tested locally that the sensors from `curl localhost:<monitoring_port>/solomon/all` are now enriched with the added attributes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ else
 DOCKER_BUILD_ARGS += --build-arg VERSION="$(OPERATOR_TAG)"
 endif
 DOCKER_BUILD_ARGS += --build-arg REVISION="$(shell git rev-parse HEAD)"
-DOCKER_BUILD_ARGS += --build-arg BUILD_DATE="$(shell date -Iseconds)"
+DOCKER_BUILD_ARGS += --build-arg BUILD_DATE="$(shell date --rfc-3339=seconds)"
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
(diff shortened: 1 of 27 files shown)


Makefile
```diff
@@ -33,7 +33,7 @@ else
 DOCKER_BUILD_ARGS += --build-arg VERSION="$(OPERATOR_TAG)"
 endif
 DOCKER_BUILD_ARGS += --build-arg REVISION="$(shell git rev-parse HEAD)"
-DOCKER_BUILD_ARGS += --build-arg BUILD_DATE="$(shell date --rfc-3339=seconds)"
+DOCKER_BUILD_ARGS += --build-arg BUILD_DATE="$(shell date -Iseconds)"
```
zlobober, line 36: Let's move this to a separate PR.

pkg/ytconfig/canondata/TestGetControllerAgentsConfig/test.canondata
```diff
@@ -4,6 +4,12 @@
         "enable_ipv6"=%false;
         retries=1000;
     };
+    "solomon_exporter"={
+        host="{POD_SHORT_HOSTNAME}";
+        "instance_tags"={
+            "k8s_pod_name"="{K8S_POD_NAME}";
```
zlobober, line 10: I'd say that "k8s_" is excessive here, maybe just "pod_name", or even "pod"? I'd prefer having the most often used labels to be as concise as possible. Also, we have "host", not "host_name"/"hostname".

Makefile
achulkov2: Ok, moved here: https://github.com/ytsaurus/yt-k8s-operator/pull/260.

pkg/ytconfig/canondata/TestGetControllerAgentsConfig/test.canondata
achulkov2: -> pod

l0kix2 (approved):

l0kix2:
LGTM, let's merge if @zlobober ok with added test.

zlobober (approved):

PR #259, opened 2024-05-06
add sidecars support to primary masters containers

kmalov:


l0kix2 (commented):
LGTM 

@koct9i could you please check that sidecars for master are consistent with sidecars for exec nodes here as intended?

koct9i:
LGTM

Next time just move this hack into common instance spec

l0kix2 (approved):

PR #261, opened 2024-05-13
Fix CRI job environment for remote exec nodes

koct9i:


pkg/components/exec_node_remote.go
```diff
@@ -54,12 +54,31 @@ func NewRemoteExecNodes(
...
+			commonSpec.ConfigOverrides,
+			map[string]ytconfig.GeneratorDescriptor{
+				consts.ContainerdConfigFileName: {
+					F: func() ([]byte, error) {
+						return cfgen.GetContainerdConfig(&spec)
+					},
+					Fmt: ytconfig.ConfigFormatToml,
+				},
+			})
+	}
```
l0kix2, line 73: Is it possible to add newBaseExecNode method and move sidecar config there so we don't copypaste it?

koct9i: Let's cleanup all that backlog later.

l0kix2 (approved):

PR #262, opened 2024-05-14
Makefile: always specify kube context

koct9i:


Makefile
```diff
@@ -193,11 +193,13 @@ kind-deploy-ytsaurus: ## Deploy sample ytsaurus cluster and all requirements.
 	$(MAKE) kind-create-cluster
 	$(MAKE) helm-kind-install
 	-$(KUBECTL) create namespace $(YTSAURUS_NAMESPACE)
+	$(KUBECTL) config set-context kind-$(KIND_CLUSTER_NAME) --namespace $(YTSAURUS_NAMESPACE)
 	$(KUBECTL) apply --server-side -n $(YTSAURUS_NAMESPACE) -f $(YTSAURUS_SPEC)
 	$(KUBECTL) wait -n $(YTSAURUS_NAMESPACE) --timeout=10m --for=jsonpath='{.status.state}=Running' --all ytsaurus
 	$(MAKE) kind-yt-info
 
 kind-undeploy-ytsaurus: ## Undeploy sample ytsaurus cluster.
+	$(KUBECTL) config use-context kind-$(KIND_CLUSTER_NAME)
```
zlobober, line 202: Why do we need this despite having `--context kind-$(KIND_CLUSTER_NAME)` in KUBECTL?

koct9i: redundant check to be sure we don't miss - it will fail if context doesn't exist

koct9i: Also I thought to rename namespace from "ytsaurus" to "ytsaurus-dev" to reduce conflict possibility.
Now name length doesn't matter because now it is default context and namespace in kubectl.

But it might confuse present users (if there are any).

l0kix2: I fully support explicitly pass the context in all commands, but I don't like that make commands silently changes k8s context for user. 
Last time it took me half an hour to figure out why my tests behaviour strange after i used this command to deploy/undeploy sample and my test were using one kind and my kubectl used another newly created one (since default one is `kind-kind`). 

If we want to make sure that context is correct lets make some assert with output `kubectl config current-context` or just be fine with `kubectl --context=<expected context>` introduced in this patch.

l0kix2 (changes requested):

l0kix2 (approved):

PR #270, opened 2024-05-31
Use ghcr.io and mirror.gcr.io in examples

savnadya:


config/samples/0.9.1/cluster_v1_minikube.yaml
```diff
@@ -0,0 +1,133 @@
+apiVersion: cluster.ytsaurus.tech/v1
```
l0kix2, line 1: maybe we should name it cluster_v1_kind.yaml or cluster_v1_local.yaml?

l0kix2 (approved):

PR #277, opened 2024-06-10
fix exec node resource calculation on non-isolated CRI-powered job environment

kruftik:


l0kix2 (approved):

pkg/components/exec_node_base.go
```diff
@@ -146,7 +146,7 @@ func (n *baseExecNode) doBuildCRISidecar(envSpec *ytv1.JobEnvironmentSpec, podSp
 	}
 
 	if n.spec.JobResources != nil {
-		jobsContainer.Resources = *n.spec.JobResources
+		jobsContainer.Resources = *n.spec.JobResources.DeepCopy()
```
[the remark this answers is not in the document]
kruftik: right, this specific field isn't changed anywhere in the code. I just 'fixed' it to be the same looking as the node resources where the bug has appeared.

pkg/components/exec_node_base_test.go
```diff
@@ -0,0 +1,320 @@
...
+	}
+}
+
+func getMasterConnectionSpec() ytv1.MasterConnectionSpec {
+	return ytv1.MasterConnectionSpec{
+		CellTag: 0,
+	}
+}
+
+func getYtsaurus() *ytv1.Ytsaurus {
```
koct9i, line 92: Please DO NOT COPY-PASTE such test assets. It will be impossible to maintain.

l0kix2: @kruft can you please fix this tests with reusing CRD building from another package?

pkg/components/exec_node_base_test.go
```diff
@@ -0,0 +1,320 @@
...
+					Protocol:      corev1.ProtocolTCP,
+				}),
+			)
+
+			bEN := baseExecNode{
+				server:        srv,
+				cfgen:         g,
+				spec:          &execNodeSpec,
+				sidecarConfig: nil,
+			}
```
l0kix2, line 310: Also this test looks rather fragile since it tests low-level implementation of the component and will often broke on refactorings. Does it really test something that couldn't be tested in ytconfig tests?

l0kix2: I've reverted this particular test file https://github.com/ytsaurus/yt-k8s-operator/pull/281
Could you please send another PR with test if you feel it is needed (I expect them to be among `ytconfig` tests and without manipulating base struct)?

PR #279, opened 2024-06-11
Add option for per location disk space quota

koct9i:
Default heuristic does not work when locations share one volume,
or when volume does not have any space limit.


l0kix2:
I've merged @kruftik PR, could you rebase please

l0kix2 (approved):

PR #282, opened 2024-06-12
CRI image cache config

koct9i:
- Fix detecting qota for volume without storage requests
  
- Add CRI image cache config


pkg/ytconfig/canondata/TestGetExecNodeConfigWithCri/single-container-without-job-resources/test.canondata
```diff
@@ -127,6 +127,9 @@
                     namespace=yt;
                     "base_cgroup"="/yt";
                 };
+                "cri_image_cache"={
+                    capacity=4294967296;
+                };
```
l0kix2, line 132: Why canondata config is changed if tests are not? Is it some default value (I don't see value in diff)?
Is it expected that config would change if no new fileds in spec is specified?

koct9i: cri image cache section is added for cri env setip

koct9i: I've added 4gb quota for image-cache location in previous pr

l0kix2: i see, so if nothing new (cache quota) would be specified in real spec, we won't have a diff in configs?

koct9i: without explicit quota there will be volume storage request.

l0kix2 (approved):

PR #286, opened 2024-06-19
[RFC] Fix https-only config

koct9i:
- Add cluster option "secureCluster"
  Issue: https://github.com/ytsaurus/yt-k8s-operator/issues/285
  
This options enforces internal clients to use secure connections and
validates that related servers are ready to handle TLS.

Also it forces TLS-only mode for native bus transport and
HTTPS-only for default role of HTTP proxies.

I.e. if it is enabled - only non-default HTTP proxies and
any RPC proxies could be not strictly TLS-only.

### DIFF
--- a/api/v1/ytsaurus_types.go
+++ b/api/v1/ytsaurus_types.go
@@ -564,6 +564,11 @@ type CommonSpec struct {
 	//+optional
 	NativeTransport *RPCTransportSpec `json:"nativeTransport,omitempty"`
 
+	// Always use TLS/HTTPS for connections inside cluster.
+	//+kubebuilder:default:=false
+	//+optional
+	SecureCluster bool `json:"secureCluster"`
+
 	// Allow prioritizing performance over data safety. Useful for tests and experiments.
 	//+kubebuilder:default:=false
 	//+optional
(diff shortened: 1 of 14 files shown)


koct9i:
Should be something like this, but this does not work yet.

"YT_PROXY=https://fqdn" is still broken somewhere.
Also ca-bundle isn't passed properly to yt cli/sdk.

sgburtsev:
> Should be something like this, but this does not work yet.
> 
> "YT_PROXY=https://fqdn" is still broken somewhere. Also ca-bundle isn't passed properly to yt cli/sdk.

JFYI: Ability to pass own CA in Go SDK was added quite a while ago: https://github.com/ytsaurus/ytsaurus/commit/16a5bebdf3e87d23213ba964b0f14412d965b86b

sgburtsev:
How will this new option `useHttps` coexist with `disableHttp`?
There would be some corner cases:
1. `useHttps` and `disableHttp` will not work without `httpsSecret`;
2. with `disableHttp` `useHttps` is mandatory.

It might make more sense to only use `disableHttp` as it almost duplicates `useHttps`.

PR #294, opened 2024-06-25
More options for store locations.

sgburtsev:
I added two additional options for the store location.
The first, `MaxTrashMilliseconds`, is intended to tune the `max_trash_ttl` parameter of the `store_locations` config. By default YTsaurus uses 1 hour.
The second one, `Watermark`, is needed to adjust location watermarks. The operator uses this proportions to calculate them:
```
storeLocation.LowWatermark = min(0.1 * storeLocation.Quota, 5GiB)
storeLocation.HighWatermark = storeLocation.LowWatermark / 2
storeLocation.DisableWritesWatermark = storeLocation.HighWatermark / 2
```
While keeping this proportions, I wanted to change the initial percentage for `LowWatermark`, which was hard-coded.
I am pretty sure there is a mistake in using `min` function. Because of `min`, it is impossible to use more than 5GiB as `LowWatermark`. However, it makes more sense to use at least 5GiB of space as `LowWatermark`. So I changed it to `max`.
Also, I added into calculations the missing parameter `TrashCleanupWatermark`.
The final solution:
```
storeLocation.LowWatermark = max(storeLocation.Watermark * storeLocation.Quota, 5GiB)
storeLocation.HighWatermark = storeLocation.LowWatermark / 2
storeLocation.DisableWritesWatermark = storeLocation.HighWatermark / 2
storeLocation.TrashCleanupWatermark = storeLocation.LowWatermark
```


config/crd/bases/cluster.ytsaurus.tech_remoteexecnodes.yaml
```diff
@@ -896,6 +901,14 @@ spec:
                       description: Disk space quota, default is size of related volume.
                       pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
                       x-kubernetes-int-or-string: true
+                    watermark:
+                      default: 10
+                      description: Location watermark threshold in percents of volume
```
zlobober, line 906: Could you please write a couple of more words (e.g. ", after exceeding which ... happens")? Even after 9 years in YT I struggle to remember what does this watermark affect, and I believe this notion is confusing for almost everybody else. 

Also, I think it is expressed in percents not of volume, bur rather of free volume.

api/v1/ytsaurus_types.go
```diff
@@ -92,6 +92,14 @@ type LocationSpec struct {
 	// Disk space quota, default is size of related volume.
 	//+optional
 	Quota *resource.Quantity `json:"quota,omitempty"`
+	// Location watermark threshold in percents of volume size.
+	//+kubebuilder:default:=10
+	//+kubebuilder:validation:Minimum:=1
+	//+kubebuilder:validation:Maximum:=50
+	Watermark int32 `json:"watermark,omitempty"`
```
l0kix2, line 99: Do I understand correctly that 10% was hardcoded and now it is the default in spec?

pkg/ytconfig/node.go
```diff
@@ -325,9 +327,11 @@ func getDataNodeServerCarcass(spec *ytv1.DataNodesSpec) (DataNodeServer, error)
 
 			// These are just simple heuristics.
 			gb := float64(1024 * 1024 * 1024)
-			storeLocation.LowWatermark = int64(math.Min(0.1*float64(storeLocation.Quota), float64(5)*gb))
+			storeLocation.LowWatermark = int64(math.Max(float64(location.Watermark)/100*float64(storeLocation.Quota), float64(5)*gb))
 			storeLocation.HighWatermark = storeLocation.LowWatermark / 2
 			storeLocation.DisableWritesWatermark = storeLocation.HighWatermark / 2
+			storeLocation.TrashCleanupWatermark = storeLocation.LowWatermark
```
l0kix2, line 333: Does it make sense to have those two fields with the same value? Is it better if we explicitly put it in existing cluster dnds configs?

api/v1/ytsaurus_types.go
sgburtsev: Yes, you are absolutely right.

l0kix2: Maybe a better approach would be to have a pointer here and in case of nil put this 10% default in _code_, that way we don't bother users with diff after updating the operator and we are free to change the default value in the future.

config/crd/bases/cluster.ytsaurus.tech_remoteexecnodes.yaml
sgburtsev: What about such description:
> Threshold, as a percentage of volume capacity, for calculation the location's low_watermark, high_watermark, disable_writes_watermark and trash_cleanup_watermark.

The abovementioned watermarks are defined in bytes in YT configs, therefore this threshold is defined as a percentage of volume capacity.
That is, `low_watermark` will be 10% of the volume capacity (by default `watermark = 10`).

I don't think this description is a right place for detailed information about each watermark. Rather it should be in the YTsaurus documentation.
This comments from code looks fine to start with:
```cpp
//! A currently full location is considered to be non-full again when available space grows
//! above this limit.
i64 LowWatermark;

//! A location is considered to be full when available space becomes less than #HighWatermark.
i64 HighWatermark;

//! All writes to the location are aborted when available space becomes less than #DisableWritesWatermark.
i64 DisableWritesWatermark;

//! Maximum amount of time files of a deleted chunk could rest in trash directory before
//! being permanently removed.
TDuration MaxTrashTtl;

//! When free space drops below this watermark, the system starts deleting files in trash directory,
//! starting from the eldest ones.
i64 TrashCleanupWatermark;
```

api/v1/ytsaurus_types.go
sgburtsev: You mean to do smth like that:
```go
if location.Watermark != nil {
   watermark = ptr.Int32(location.Watermark.Value());
} else {
   watermark = 10;
}
```
In fact, I intentionally made explicit default to make its value transparent for the users. And we should bother them with diff in case of changing the default as it affects the watermarks. I definitely want to be bothered by such change 😄 .
For me, frankly saying, it's much better to see explicit defaults rather than to search them in the source code of YTsaurus for all optional fields. But at least here we have to keep the default since this parameter is operator specific only.

l0kix2: I think, that in the ideal world, default should be only mentioned in the docs (+release notes if changed) and the user shouldn't have it in his spec and see the diff and so on. Although maybe you're right, and some users don't want to read release notes (or we can forget to mention some stuff there) and prefer to see diff in the spec. I'll discuss that with other maintainers.

In this particular case it is kinda confusing for user to see the diff if in reality the value was not changed.

This is minor thing anyway since we don't have some policy here, you can left it as it is, we explain diff in the release notes.

l0kix2 (approved):
LGTM

@zlobober can you validate (or tag someone who can) that:
 - low watermark should be `max` of the specified percentage of quota and 5Gib and not `min`
 - it is make sense to explicitly set `trash_cleanup_watermark` equal to `low_watermark` for all users
?

config/crd/bases/cluster.ytsaurus.tech_remoteexecnodes.yaml
sgburtsev: I changed the description to a more understandable one, but to fit into the 80 character limit.

sgburtsev:
After discussion with @psushin I changed the initial proposal in the watermarks calculation part. Now it looks like that:
```
storeLocation.LowWatermark = min(0.1 * storeLocation.Quota, 25GiB)
storeLocation.HighWatermark = storeLocation.LowWatermark / 2
storeLocation.DisableWritesWatermark = storeLocation.HighWatermark / 2
storeLocation.TrashCleanupWatermark = storeLocation.LowWatermark
```
with ability to provide your own LowWatermark.
Such approach should work well in both small testing environments and production with high-capacity volumes .

pkg/ytconfig/node.go
```diff
@@ -323,11 +325,18 @@ func getDataNodeServerCarcass(spec *ytv1.DataNodesSpec) (DataNodeServer, error)
 		if quota != nil {
 			storeLocation.Quota = *quota
 
+			if location.LowWatermark != nil {
+				storeLocation.LowWatermark = location.LowWatermark.Value()
+			} else {
+				gb := float64(1024 * 1024 * 1024)
+				storeLocation.LowWatermark = int64(math.Min(0.1*float64(storeLocation.Quota), float64(25)*gb))
+			}
```
psushin, line 333: We discussed it in DM, this looks as a reasonable default to me

IMO, all the further fine tuning should be done via dynamic config.

l0kix2 (approved):

zlobober:
I'm ok with this change.

sgburtsev:
> regenerate yamls please so linter don't complain

Committed needed changes. Sorry for that.

PR #296, opened 2024-06-26
Remove mutating/defaulter webhooks

koct9i:


api/v1/ytsaurus_webhook.go
```diff
@@ -45,36 +47,6 @@ func (r *Ytsaurus) SetupWebhookWithManager(mgr ctrl.Manager) error {
...
-							consts.YTComponentLabelName: fmt.Sprintf("%s-%s", r.Name, "yt-master"),
-						},
-					},
-					TopologyKey: "kubernetes.io/hostname",
-				},
-			},
-		}
-	}
-}
-
```
l0kix2, line 77: Do this affinities not needed or set somewhere else?

l0kix2 (commented):
Will defaults be applied at all without hook?

api/v1/ytsaurus_webhook.go
koct9i: Not always needed. I've added validation instead.

koct9i:
> Will defaults be applied at all without hook?

Yep. Defaults should be applied on read automatically.

https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#defaulting

l0kix2 (approved):

PR #305, opened 2024-07-02
added validateExistsYTsaurus function for validation webhook

### LINKED ISSUE #241: Protection from deploying more than 1 ytsaurus in namespace
Currently names of pods/sts do not have cluster names in them, so deploying second ytsaurus in namespace would be a problem.
We want some protection from that. 
Maybe validation hook, maybe in reconciler itself.

qurname2:
To fix #241


api/v1/ytsaurus_webhook.go
```diff
@@ -489,25 +532,24 @@ func (r *Ytsaurus) evaluateYtsaurusValidation(old *Ytsaurus) error {
 }
 
 // ValidateCreate implements webhook.Validator so a webhook will be registered for the type
-func (r *Ytsaurus) ValidateCreate() (admission.Warnings, error) {
-	ytsauruslog.Info("validate create", "name", r.Name)
-
-	return nil, r.evaluateYtsaurusValidation(nil)
+func (r *ytsaurusValidator) ValidateCreate(ctx context.Context, _ runtime.Object) (admission.Warnings, error) {
+	ytsauruslog.Info("validate create", "name", r.Ytsaurus.Name)
+	return nil, r.Ytsaurus.evaluateYtsaurusValidation(ctx, nil)
```
l0kix2, line 537: The reason webhook tests are failing is that before `ValidateCreate` was called on ytsaurus resource
https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/webhook/admission/validator.go#L92

but with a custom validator resource is being passed as an argument instead (suppose it is the same object)
https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/webhook/admission/validator_custom.go#L91

So now `r` is empty struct and all validate methods must be fixed to explicitly receive validated object and old object (in update, in create it is nil) (and I suppose it makes sense to move all the methods here to a new ytsaurusValidator struct).

api/v1/ytsaurus_webhook.go
```diff
@@ -451,7 +463,37 @@ func (r *Ytsaurus) validateInstanceSpec(instanceSpec InstanceSpec, path *field.P
...
+	}
+	k8sClient, err := client.New(cfg, client.Options{Scheme: scheme.Scheme})
+	if err != nil {
+		allErrors = append(allErrors, field.InternalError(field.NewPath("k8s client"), err))
+		return allErrors
+	}
+
+	var ytsaurusList YtsaurusList
+
+	err = k8sClient.List(ctx, &ytsaurusList, &client.ListOptions{Namespace: ytsaurus.Namespace})
```
l0kix2, line 487: Also as far as I remember this k8sClient gives an error and all point of introducing custom validator is to use client which is came from the manager here.

l0kix2 (changes requested):

api/v1/ytsaurus_webhook.go
l0kix2: I suggest start by commenting other validators and make your new test with new validator (and kubeclient from validator struct) green. So we don't discover k8s call still don't work after fixing other validators.

api/v1/ytsaurus_webhook.go
```diff
@@ -451,63 +460,102 @@ func (r *Ytsaurus) validateInstanceSpec(instanceSpec InstanceSpec, path *field.P
...
+	err := r.Client.List(ctx, &ytsaurusList, &client.ListOptions{Namespace: newYtsaurus.Namespace})
+	ytsauruslog.Info("validateExistsYTsaurus", "ytsaurusList", ytsaurusList)
+
+	if err != nil && !apierrors.IsNotFound(err) {
+		allErrors = append(allErrors, field.InternalError(field.NewPath("k8sClient"), err))
+		return allErrors
+	}
+
+	for _, ytsaurus := range ytsaurusList.Items {
+		if ytsaurus.Name == newYtsaurus.Name {
```
koct9i, line 477: So. Validation will accept any update when namespace already has more than one ytsaurus instance.
This is fine, but we could do better.
For example: accept only updates to new spec with isManaged = false.

api/v1/ytsaurus_webhook.go
```diff
@@ -41,48 +42,56 @@ import (
 // log is for logging in this package.
 var ytsauruslog = logf.Log.WithName("ytsaurus-resource")
 
+type ytsaurusValidator struct {
+	Scheme *runtime.Scheme
```
koct9i, line 46: Pointer to Scheme seems unused

l0kix2: hm, why linters haven't catched that

l0kix2:
LGTM, tests ok, let's fix issues highlighted by @koct9i and merge.

api/v1/ytsaurus_webhook.go
qurname2: Fixed

l0kix2 (approved):

api/v1/ytsaurus_webhook.go
qurname2: Improved logic in that place, please take a look https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/305/commits/c328594faf54c70202eb093de5caf7dbad5793d3

koct9i (approved):

PR #310, opened 2024-07-11
Use `secure` instead of `useInsecureCookies`, pass caBundle to UI

sgburtsev:
This change allows to use HTTPS connection from the UI to the cluster's http-proxies.

1. `secure` option is passed to the UI config.
2. `caBundle` is passed to the UI pods.
3. The webhook validates the presence of `httpsSecret` for the default proxy in case of a secure connection.

### DIFF
--- a/api/v1/ytsaurus_types.go
+++ b/api/v1/ytsaurus_types.go
@@ -479,6 +479,10 @@ type UISpec struct {
 	//+kubebuilder:default:=NodePort
 	ServiceType  corev1.ServiceType `json:"serviceType,omitempty"`
 	HttpNodePort *int32             `json:"httpNodePort,omitempty"`
+	// If defined allows insecure (over http) authentication.
+	// Deprecated: use `secure` instead.
+	//+optional
+	UseInsecureCookies *bool `json:"useInsecureCookies"`
 	// Use secure connection to the cluster's http-proxies.
 	//+kubebuilder:default:=false
 	//+optional
--- a/api/v1/ytsaurus_types.go
+++ b/api/v1/ytsaurus_types.go
@@ -479,11 +479,12 @@ type UISpec struct {
 	//+kubebuilder:default:=NodePort
 	ServiceType  corev1.ServiceType `json:"serviceType,omitempty"`
 	HttpNodePort *int32             `json:"httpNodePort,omitempty"`
-	//+kubebuilder:default:=true
+	// Use secure connection to the cluster's http-proxies.
+	//+kubebuilder:default:=false
 	//+optional
-	UseInsecureCookies bool                        `json:"useInsecureCookies"`
-	Resources          corev1.ResourceRequirements `json:"resources,omitempty"`
-	InstanceCount      int32                       `json:"instanceCount,omitempty"`
+	Secure        bool                        `json:"secure"`
+	Resources     corev1.ResourceRequirements `json:"resources,omitempty"`
+	InstanceCount int32                       `json:"instanceCount,omitempty"`
 
 	// If defined it will be used for direct heavy url/commands like: read_table, write_table, etc.
 	//+optional
--- a/api/v1/ytsaurus_webhook.go
+++ b/api/v1/ytsaurus_webhook.go
@@ -398,6 +398,26 @@ func (r *Ytsaurus) validateYQLAgents(*Ytsaurus) field.ErrorList {
 	return allErrors
 }
 
+func (r *Ytsaurus) validateUi(*Ytsaurus) field.ErrorList {
+	var allErrors field.ErrorList
+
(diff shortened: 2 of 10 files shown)


api/v1/ytsaurus_types.go
```diff
@@ -479,11 +479,12 @@ type UISpec struct {
 	//+kubebuilder:default:=NodePort
 	ServiceType  corev1.ServiceType `json:"serviceType,omitempty"`
 	HttpNodePort *int32             `json:"httpNodePort,omitempty"`
-	//+kubebuilder:default:=true
+	// Use secure connection to the cluster's http-proxies.
+	//+kubebuilder:default:=false
 	//+optional
-	UseInsecureCookies bool                        `json:"useInsecureCookies"`
```
l0kix2, line 484: Let's not delete but deprecate the field, giving users the opportunity to migrate.
I suggest we add validation that values don't conflict (smth like `UseInsecureCookies == !Secure`) and after that use `Secure` field in the code as you did it.

l0kix2 (changes requested):

l0kix2:
@koct9i please look at this PR so if merged it wound conflict with https-changes you're working on in #286

api/v1/ytsaurus_types.go
koct9i: Yeah. This seems reasonable way of deperecation.
Remove default=true and convert into *bool - to not trigger validation if this field isn't here.

I hope validation is called before each reconcile. If not we should do so.

koct9i:
> @koct9i please look at this PR so if merged it wound conflict with https-changes you're working on in #286

Should be fine. go on

api/v1/ytsaurus_types.go
sgburtsev: `UseInsecureCookies = false` did not make any sense before this change. So I added check for ambiguous `UseInsecureCookies == false && Secure == false`, ignoring `UseInsecureCookies == true`.

l0kix2: > UseInsecureCookies = false did not make any sense before this change
I don't get why it didn't make sense. useInsecureCookies was true by default and `YT_AUTH_ALLOW_INSECURE` was set in this case, but if user wanted to disallow using insecure cookies — he might set false and expect UI to use secure cookie.

So now we are making this nil+optional and should think about 3 cases:
1) users who didn't have useInsecureCookies set in their spec — default behaviour with insecure cookie should work for them (which is work since secure is false by default)
2) users who explicitely set useInsecureCookies=true in their spec — they should be fine since we check secure which is false by default and ignore the old field.
3) users who explicitly set useInsecureCookies=false — their spec will be broken after operator update, so we must defend them and add validation (which is what you did).

So I think current code shouldn't break anyone.

pkg/components/ui.go
```diff
@@ -45,6 +46,11 @@ func NewUI(cfgen *ytconfig.Generator, ytsaurus *apiproxy.Ytsaurus, master Compon
 		image = *resource.Spec.UI.Image
 	}
 
+	var caBundle *resources.CABundle
+	if caBundleSpec := resource.Spec.CABundle; caBundleSpec != nil {
+		caBundle = resources.NewCABundle(caBundleSpec.Name, consts.CABundleVolumeName, consts.CABundleMountPoint)
+	}
```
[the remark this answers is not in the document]
sgburtsev: It's impossible to use an HTTPS connection with self-signed certificates without CABundle. That's why it's an essential part of the PR.

api/v1/ytsaurus_webhook.go
```diff
@@ -398,6 +398,31 @@ func (r *Ytsaurus) validateYQLAgents(*Ytsaurus) field.ErrorList {
...
+						field.NewPath("spec", "httpProxies").Index(i).Child("transport", "httpsSecret"),
+						fmt.Sprintf("configured HTTPS for proxy with `%s` role is required for ui.secure", consts.DefaultHTTPProxyRole)))
+				}
+				break
+			}
+		}
+	}
+
+	return allErrors
+}
```
l0kix2, line 424: On the second thought, I think we break some user experience here. 
Before the change user could choose if he wants https or http cookie for frontend (it being set via YT_AUTH_ALLOW_INSECURE) and after the change user can either use http-cookie or if he wants https-cookie he have to configure https for proxies.
And this is two different things which are now depend on one setting.

I suggest we un-deprecate `UseInsecureCookies` settings and let it configure cookies and `Secure` would configure https for UI <-> proxy communication.

@koct9i @sgburtsev am I missing something?

sgburtsev: Oh, I see. You are right. I'll revert it in a separate PR.

PR #313, opened 2024-07-16
Add nodeSelector, tolerations, affinity, strategy, issuerRef to the Helm chart

sgburtsev:
While `nodeSelector` was supported by `helmify`, other needed fields are not. Thus, ugly sed commands are needed to generate the chart.

Considering this change as a temporary hack, I see this options:
1. Create an issue/PR with support for needed fields in the `helmify` project.
2. Stop generating the chart. Maintain it manually, only copying generated CRDs.
3. Generate the chart with other tools. For instance, https://github.com/yeahdongcn/kustohelmize.
4. Write own tool 😎.


l0kix2:
@koct9i do you have ideas how to implement this better?

l0kix2:
> Stop generating the chart. Maintain it manually, only copying generated CRDs.

We've discussed a little and this one looks like a best approach for now.
I suggest to do:
1) PR1 
— remove helmify calls from the makefile (chart stays as is) 
— add `make helmify-crd` (or just `make helmify`) which only will copy kubebuilder generated crds
— add `make helmify-crd` into the `make lint-generated` target to check if someone forgot to copy crds in chart
2) PR2
— add required changes to (already not helmify-managed) chart.

sgburtsev:
> Let me check other options. Do you have experience with kustohelmize — does it have what needed from the box?

I haven't tried it, but from the first glance it has much more customization. It also has some drawbacks. For instance, the names of generated files are in form apigroup_kind_name.yaml.

PR #320, opened 2024-08-02
Static helm

sgburtsev:
As suggested in https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/313#issuecomment-2242865245 I got rid of the helmify step.

The resulting CRDs are almost identical, except for the wrapping of long lines and missing labels (which are not necessary).


Makefile
```diff
@@ -410,6 +407,16 @@ kind: $(KIND) ## Download kind locally if necessary.
 $(KIND): $(LOCALBIN)
 	$(call go-install-tool,$(KIND),sigs.k8s.io/kind,$(KIND_VERSION))
 
+.PHONY: envsubst
+envsubst: $(ENVSUBST) ## Download envsubst locally if necessary.
+$(ENVSUBST): $(LOCALBIN)
+	$(call go-install-tool,$(ENVSUBST),github.com/a8m/envsubst/cmd/envsubst,$(ENVSUBST_VERSION))
+
+.PHONY: kubectl-slice
+kubectl-slice: $(KUBECTL_SLICE) ## Download yq locally if necessary.
```
l0kix2, line 416: yq --> kubectl-slice

config/crd/patches/cainjection_in_chyts.yaml
```diff
@@ -3,5 +3,5 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    cert-manager.io/inject-ca-from: $(CERTIFICATE_NAMESPACE)/$(CERTIFICATE_NAME)
+    cert-manager.io/inject-ca-from: $(WEBHOOK_CERTIFICATE_NAMESPACE)/$(WEBHOOK_CERTIFICATE_NAME)
```
l0kix2, line 6: Why is it changed, are we using wrong variables currently?

l0kix2 (approved):
Looks good, helm chart looks almost the same. 
I only not sure why those `CERTIFICATE_NAMESPACE` variables are replaced

config/crd/patches/cainjection_in_chyts.yaml
sgburtsev: Yes, there was a mistake. That's why previously the annotation looked like:
```
cert-manager.io/inject-ca-from: '{{ .Release.Namespace }}/{{ include "ytop-chart.fullname" . }}-$(CERTIFICATE_NAMESPACE)/$(CERTIFICATE_NAME)'
```

wilwell:
@sgburtsev, it looks like this change has broken the Helm chart construction.

If you [look](https://hub.docker.com/r/ytsaurus/ytop-chart-nightly/tags) at the Helm chart created before and after your change, it's clear that your change has affected the generation of multiple YAML files.

For example, `deployment.yaml` is missing now. It only generates `templates/crds` files.

Could you please take a look? Maybe you understand why your change triggered this behavior?

wilwell:
I've fixed this issue in 5863668ed96723dba96eff8b2ed81dd001ea3dda, **but there is still one more bug.**

In the files values.yaml the image tag of `ytsaurus/k8s-operator` is always `0.0.0-alpha`, even if you redefine `RELEASE_VERSION` with another value.

@sgburtsev do you know how to fix that?

sgburtsev:
Unfortunately, I didn't take into account github workflows after this changes.
I will take a look, but the main purpose of the PR was exactly what you said: to generate only CRDs and leave other as static.

wilwell:
How to reproduce the behaviour:

In make file redefine release version, e.g. `RELEASE_VERSION = "1.1.1"`

run 2 commands from `release` in Makefile:
```
$(MAKE) helm-chart
helm package $(OPERATOR_CHART)
```

If you unpack generated helm package, you find that image tag of `ytsaurus/k8s-operator` is `0.0.0-alpha`.

------------

Moreover: if you remove tag: `0.0.0-alpha` from `values.yaml` in `ytop-chart` from repo and then regenerate helm package, you will find that tage field is absent in the generated package too.

I suppose that new helm-chart command just copies `values.yaml` from `ytop-chart` in repo (!?)

sgburtsev:
It's quite a discovery for me that nightly build has another name. I supposed that a nightly and a release build should be completely interchangeable. @l0kix2, do you know the reason for that?
As for the tag for releases, it could be easily fixed without need of generating values. I will prepare a fix ASAP.
The motivation of moving from a generated Chart to a static one is in #313.

l0kix2:
> It's quite a discovery for me that nightly build has another name. I supposed that a nightly and a release build should be completely interchangeable. @l0kix2, do you know the reason for that?

I don't know the answer. I don't think we expect them to be different in any way, so I would suggest we have one image name for operator and helm chart with either release (like 0.0.15) or dev (like 0.0.428-dev-5863668ed96723dba96eff8b2ed81dd001ea3dda) version part.

@savnadya do we have a good reason to have different names here?

PR #324, opened 2024-08-14
add forceTcp and keepSocket options

leo-astorsky:


api/v1/ytsaurus_types.go
```diff
@@ -571,6 +571,13 @@ type CommonSpec struct {
 	//+kubebuilder:default:=false
 	//+optional
 	UseIPv4 bool `json:"useIpv4"`
+	//+kubebuilder:default:=true
+	//+optional
+	KeepSocket bool `json:"keepSocket"`
+	//+kubebuilder:default:=false
+	//+optional
+	ForceTcp bool `json:"forceTcp"`
+
```
leo-astorsky, line 580: force_tcp - false
keep_socket - true

cause of https://github.com/ytsaurus/ytsaurus/blob/b41294f14e346c455b1d69d268b029a5d07f71f5/yt/yt/core/dns/config.cpp#L21-L24

pkg/ytconfig/canondata/TestGetStrawberryControllerConfig/test.canondata
```diff
@@ -11,7 +11,9 @@
         chyt={
             "address_resolver"={
                 "enable_ipv4"=%true;
-                "enable_ipv6"=%true;
+                "enable_ipv6"=%false;
```
leo-astorsky, line 14: cause of
- https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/324/files#diff-9875b847729a36cdf9f8ee8333cefa406ed3e7fb1f5f297c87e5e3e556d907bbR558
- https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/324/files#diff-ee85a92ac74a0dc4dbb84dab8829fe475be259f340407f523529d5dda26961ddR315-R317

l0kix2:
Thanks, will look into it soon.

api/v1/ytsaurus_types.go
```diff
@@ -571,6 +571,13 @@ type CommonSpec struct {
 	//+kubebuilder:default:=false
 	//+optional
 	UseIPv4 bool `json:"useIpv4"`
+	//+kubebuilder:default:=true
+	//+optional
+	KeepSocket bool `json:"keepSocket"`
+	//+kubebuilder:default:=false
+	//+optional
+	ForceTcp bool `json:"forceTcp"`
```
l0kix2, line 579: I think better call it `ForceTCP` (and in tag also) according to go [style guides ](https://go.dev/wiki/CodeReviewComments#initialisms) similarly to HTTP even if linter doesn't triggered on that.

pkg/ytconfig/common.go
```diff
@@ -51,6 +51,8 @@ type ClusterConnection struct {
 type AddressResolver struct {
 	EnableIPv4 bool `yson:"enable_ipv4"`
 	EnableIPv6 bool `yson:"enable_ipv6"`
+	KeepSocket bool `yson:"keep_socket"`
+	ForceTcp   bool `yson:"force_tcp"`
```
l0kix2, line 55: I suggest (despite existing `EnableIP*` fields) we go with `*bool` + `omitempty` (and in the spec struct also) to avoid hardcoding ytsaurus defaults in the operator, just giving the users an option to override the defaults. 
That way it is expected we don't have a diff in our canondata (and add extra test where we are setting new fields). Also users won't have to do full update with a new release of the operator.

l0kix2 (changes requested):

api/v1/ytsaurus_types.go
leo-astorsky: > ForceTCP

```bash
$ make lint
api/v1/ytsaurus_types.go:579:17: json(camel): got 'forceTCP' want 'forceTcp' (tagliatelle)
        ForceTCP *bool `json:"forceTCP,omitempty"`
                       ^
make: *** [lint] Error 1
```


add like - 
```
	//+kubebuilder:default:=false
	//+optional
	//nolint:tagliatelle
	ForceTCP *bool `json:"forceTCP,omitempty"`
```
?

l0kix2: Let's just stick with what linter wants in json tag (`forceTcp `) then.

We might configure tagliatelle  linter with `goCamel` instead of `camel`, but I suppose it will want other fields renamed which is not backward compatible.

pkg/ytconfig/common.go
leo-astorsky: How to do it in crds?
- https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/324/files#diff-d4dd7e778063446c32712d7adbff415c715fe87ce7607b3762da91d51f8fe9acR10038-R10040
- https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/324/files#diff-d4dd7e778063446c32712d7adbff415c715fe87ce7607b3762da91d51f8fe9acR12512-R12514

l0kix2: I haven't understood your question. You use pointer in go struct and crds are generated automatically with `make generate`.

leo-astorsky:
```
`make manifests` generate:
              forceTcp:
                default: false
                type: boolean
```
And the code (operator) will then generate the full config with force_tcp (even if i not define in my .values).
But not for tests.

Aka,
- "That way it is expected we don't have a diff in our canondata" - Yes
- "Also users won't have to do full update with a new release of the operator" - No, they will have a full update :(

leo-astorsky: ah, i think - `+kubebuilder:default:=None`  - is my option

l0kix2: > "Also users won't have to do full update with a new release of the operator" - No, they will have a full update :( (cause crd's defaults will generate fullspec with force_tcp and keep_socket default values)

For users it doesn't matter if crd has new fields, I'm saying that operator will recreate all the components if it sees any diff in components static config. But if [here]("Also users won't have to do full update with a new release of the operator" - No, they will have a full update :( (cause crd's defaults will generate fullspec with force_tcp and keep_socket default values)) you use pointers with `omitempty` there will be no chagens in static configs and operator wouldn't bother users with recreating their yt cluster without need.

l0kix2: > ah, i think - +kubebuilder:default:=None - is my option, sorry, forgot to edit it

Not quite sure with this defaults, maybe we can just not add `kubebuilder:default` comment here and it will be fine

leo-astorsky: > maybe we can just not add kubebuilder:default comment here and it will be fine

Yes, this solve my questions above.
Sent a new commit.

pkg/ytconfig/canondata/TestGetStrawberryControllerConfig/test.canondata
l0kix2: So it was hardcoded for strawberry, but now it is configured according to the common value, ok

pkg/ytconfig/generator_test.go
```diff
@@ -402,6 +402,18 @@ func TestGetYQLAgentConfig(t *testing.T) {
 	canonize.Assert(t, cfg)
 }
 
+func TestResolverOptionsKeepSocketAndForceTCP(t *testing.T) {
+	ytsaurus := getYtsaurusWithEverything()
+	forceTCP := true
+	keepSocket := true
+	ytsaurus.Spec.CommonSpec.ForceTCP = &forceTCP
+	ytsaurus.Spec.CommonSpec.KeepSocket = &keepSocket
```
l0kix2, line 410: `ptr.To(..)` can be used here for brevity, but it is minor.

pkg/ytconfig/generator.go
```diff
@@ -310,7 +312,9 @@ func (g *Generator) GetClusterConnection() ([]byte, error) {
 }
 
 func (g *Generator) GetStrawberryControllerConfig() ([]byte, error) {
-	c := getStrawberryController()
+	var resolver AddressResolver
+	g.fillAddressResolver(&resolver)
+	c := getStrawberryController(&resolver)
```
l0kix2, line 317: Let's pass not pointer but struct to `getStrawberryController` since we are not changing it inside that function and nil value there is unexpected and will cause panic, so safer and more straightforward is just to just pass `resolver` value here.

l0kix2 (changes requested):
Looks good overall, I only have a couple of minor comments.

pkg/ytconfig/canondata/TestGetStrawberryControllerConfig/test.canondata
l0kix2: On a second though that may be backward incompatible change and maybe we need to think here more.

l0kix2: So users, who have `ipv4=false;ipv6=true` or `ipv4=true;ipv6=false` (assume false/false is invalid configuration) will redeploy strawberry on operator update with this settings, not true/true as before which can break things for them. Not sure yet how problematic such update could be.

pkg/ytconfig/chyt.go
```diff
@@ -27,7 +27,18 @@ type ChytInitCluster struct {
...
+	AddressResolver AddressResolver `yson:"address_resolver"`
+}
+
+func getStrawberryController(resolver *AddressResolver) StrawberryController {
+	chytConfig := ChytConfig{
+		AddressResolver: *resolver,
+	}
+	chytYsonConfig, err := marshallYsonConfig(chytConfig)
+	if err != nil {
+		chytYsonConfig = yson.RawValue("{address_resolver={enable_ipv4=%true;enable_ipv6=%true;retries=1000}}")
```
achulkov2, line 40: I don't like this, why would the provided address resolver config be erroneous in the first place?
If it can be null let's handle that explicitly, otherwise there should be no errors. 

Also I believe we are losing `retries=1000`. Let's keep it for now in some way.

achulkov2 (changes requested):

pkg/ytconfig/canondata/TestGetStrawberryControllerConfig/test.canondata
l0kix2: I suppose we can describe breaking changes in the next release about strawberry ipv4/6.

pkg/ytconfig/chyt.go
l0kix2: Yes, I agree on silent error stuff. Let's just return error from this function.

l0kix2: I think retries are ok, since they are hardcoded in fillAddressResolver (and we don't see diff in canondata)

leo-astorsky:
Sent a commit with fixes

l0kix2 (approved):

PR #326, opened 2024-08-26
Refactor label names for components

achulkov2:
Example for master (single group of instances):
``` 
labels:
    app.kubernetes.io/component: yt-master
    app.kubernetes.io/instance: minisaurus-yt-master
    app.kubernetes.io/managed-by: ytsaurus-k8s-operator
    app.kubernetes.io/name: yt-master
    app.kubernetes.io/part-of: ytsaurus
    apps.kubernetes.io/pod-index: "0"
    controller-revision-hash: ms-78dc745cc4
    statefulset.kubernetes.io/pod-name: ms-0
    yt_cluster: minisaurus
    yt_component_instance: minisaurus-yt-master
    yt_component_type: yt-master
```
Example for http proxy (different roles):
```
labels:
    app.kubernetes.io/component: yt-http-proxy-control
    app.kubernetes.io/instance: minisaurus-yt-http-proxy-control
    app.kubernetes.io/managed-by: ytsaurus-k8s-operator
    app.kubernetes.io/name: yt-http-proxy
    app.kubernetes.io/part-of: ytsaurus
    apps.kubernetes.io/pod-index: "0"
    controller-revision-hash: hp-control-8675dc567
    statefulset.kubernetes.io/pod-name: hp-control-0
    yt_cluster: minisaurus
    yt_component_instance: minisaurus-yt-http-proxy-control
    yt_component_type: yt-http-proxy
```
Docs about standard labels: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels.
Primary rationale: we need some labels that identify a component as a whole, across all racks/roles/etc. This is useful for monitoring, for example.


kruftik:
Jfyi, label selector is an immutable field in Sts/deployment and other pod controller resource kinds. It can't be changed without resource deletion. For now if I am not mistaken, yt_component is used as the only lable selector. I'd prefer not to brake things here and leave this label as well

achulkov2:
> Jfyi, label selector is an immutable field in Sts/deployment and other pod controller resource kinds. It can't be changed without resource deletion. For now if I am not mistaken, yt_component is used as the only lable selector. I'd prefer not to brake things here and leave this label as well

Ok, reverted to the old name. It is a bit confusing, but oh well.

koct9i:
https://kubernetes.io/docs/reference/labels-annotations-taints/

koct9i:
I suggest this hierarchy
```
app.kubernetes.io/managed-by: ytsaurus-k8s-operator
app.kubernetes.io/part-of: ytsaurus-{cluster_name}
app.kubernetes.io/compontent: ytsaurus-{component_type}
app.kubernetes.io/name: ytsaurus-{component_type}-{instance_group}
ytsaurus.tech/cluster-name: {cluster_name}
```

These are labels for stateful-sets - "instance" isn't unique at this level.

And keep "yt_component" as is, and remove it later.

achulkov2:
> I suggest this hierarchy
> 
> ```
> app.kubernetes.io/managed-by: ytsaurus-k8s-operator
> app.kubernetes.io/part-of: ytsaurus-{cluster_name}
> app.kubernetes.io/name: ytsaurus-{component_type}
> app.kubernetes.io/compontent: ytsaurus-{component_type}-{instance_group}
> ytsaurus.tech/cluster-name: {cluster_name}
> ```
> 
> These are labels for stateful-sets - "instance" isn't unique at this level.
> 
> And keep "yt_component" as is, and remove it later.
> 
> UPD: swapped "name" and "component" - name is used as monitoring bucket.

Implemented this, but kept `yt` instead of `ytsaurus` for now, so that we do not have to change all the values of existing labels.

pkg/labeller/labeller.go
```diff
@@ -18,6 +18,7 @@ type FetchableObject struct {
 type Labeller struct {
 	APIProxy       apiproxy.APIProxy
 	ObjectMeta     *metav1.ObjectMeta
+	ComponentType  string
```
l0kix2, line 21: It is rather confusing that you are adding ComponentType field and it doesn't have `consts.ComponentType` type and content of those consts.

Maybe we could reduce complexity instead of adding it? 
I think it would be simpler if Labeller would look something like:
```
type Labeller struct {
	apiProxy       apiproxy.APIProxy
	objectMeta     *metav1.ObjectMeta
        componentType const.ComponentType
        componentNamePart string // role for hp, name for others, may be empty
        annotations    map[string]string
}
```
and some methods over those fields to get full label and full name.
internal map const.ComponentType <-> YTComponentLabel* may be introduced to reduce possible errors and simplify labeller usage.

I would prefer to have some tests with canonized labels here before refactoring, though I'm not sure if it is easy to write since labeller is configured inside those components and currently the easiest way looks like to fetch such labels from k8s resources after applying.

achulkov2: Yeah, the annoying part is that it is an even bigger refactoring, since ComponentType and ComponentLabel are almost the same (except for CamelCase vs lower-case-hyphenated), but slightly different. 

I guess the best way for now is indeed to have an internal function which maps these types 1:1 without changing the names.

achulkov2: As for tests, maybe I can try fetching these labels in an e2e test. Not sure I want to do this for all components though.

l0kix2: > Yeah, the annoying part is that it is an even bigger refactoring

I see it something like:
- `ComponentLabel` usage are auto-replaced with a new `ComponentLabel()` call
- `ComponentName` usages are replaced with a new `FullComponentName()` call
- Labeller initialization is replaced with `NewLabeller(..., componentType ComponentType, componentNamePart string //may be empty)`
- `ComponentType()` and other methods for labels are added.
Do you see more things to be done?

pkg/labeller/labeller.go
```diff
@@ -88,11 +108,26 @@ func (l *Labeller) GetListOptions() []client.ListOption {
...
+		"app.kubernetes.io/part-of": l.GetPartOfLabelValue(),
+		// Uppercase looks awful, even though it is more typical for k8s.
+		"app.kubernetes.io/managed-by": "ytsaurus-k8s-operator",
+		// It is nice to have the cluster name as a label.
+		// Template: <cluster_name>.
+		consts.YTClusterLabelName: l.GetClusterName(),
+		// This label is used to check pods for readiness during updates.
+		// The name isn't quite right, but we keep it for backwards compatibility.
+		// Template: <cluster_name>-yt-<component_type>-<instance_group>[-init-job].
+		consts.YTComponentLabelName: l.GetInstanceLabelValue(isInitJob),
```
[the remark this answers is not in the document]
achulkov2: How will we know when to deprecate `instance`? Right now it is equal to the name of the YT cluster, which is extremely weird, IMO.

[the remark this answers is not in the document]
achulkov2: I would still remove it.

pkg/labeller/labeller.go
achulkov2: Implemented this ~as you said. Had to get rid of `DeprecatdeChytController` for simplicity, but I think that's fine.

l0kix2 (approved):
Nice, and with tests!

l0kix2:
We will need to add some details to release notes of labels and deprecated chyt field also.

PR #328, opened 2024-08-28
Added tolerations for Strawberry

qurname2:
To resolve https://github.com/ytsaurus/ytsaurus-k8s-operator/issues/327


pkg/components/microservice.go
```diff
@@ -129,6 +133,7 @@ func (m *microserviceImpl) rebuildDeployment() *appsv1.Deployment {
 			Image: m.image,
 		},
 	}
+	m.builtDeployment.Spec.Template.Spec.Tolerations = m.tolerations
```
l0kix2, line 136: Aren't they already being set in `deployment.Build()`?

pkg/components/strawberry_controller.go
```diff
@@ -221,6 +226,7 @@ func (c *StrawberryController) syncComponents(ctx context.Context) (err error) {
 			VolumeMounts: volumeMounts,
 		},
 	}
+	deployment.Spec.Template.Spec.Tolerations = c.microservice.getTolerations()
```
l0kix2, line 229: Aren't they being set in `microservice.buildDeployment()` ?

pkg/components/microservice.go
qurname2: Yes, looks like you right, overengineered that a bit =)

pkg/components/strawberry_controller.go
qurname2: Yes, looks like you right, overengineered that a bit =)

pkg/components/microservice.go
```diff
@@ -46,6 +46,7 @@ type microserviceImpl struct {
 	builtDeployment *appsv1.Deployment
 	builtService    *corev1.Service
 	builtConfig     *corev1.ConfigMap
+	tolerations     []corev1.Toleration
```
l0kix2, line 49: It looks like you don't need it in this struct

pkg/components/microservice.go
```diff
@@ -54,12 +55,14 @@ func newMicroservice(
...
+	tolerations []corev1.Toleration) microservice {
 	return &microserviceImpl{
 		labeller:      labeller,
 		image:         image,
 		instanceCount: instanceCount,
 		ytsaurus:      ytsaurus,
+		tolerations:   tolerations,
 		service: resources.NewHTTPService(
 			serviceName,
 			nil,
```
l0kix2, line 68: but need to pass tolerations in NewDeployment instead.

l0kix2 (changes requested):

pkg/components/strawberry_controller.go
```diff
@@ -64,6 +65,9 @@ func NewStrawberryController(
 		if resource.Spec.StrawberryController.Image != nil {
 			image = *resource.Spec.StrawberryController.Image
 		}
+		if len(resource.Spec.StrawberryController.Tolerations) != 0 {
+			tolerations = resource.Spec.StrawberryController.Tolerations
+		}
```
l0kix2, line 70: No point of this check, just pass `resource.Spec.StrawberryController.Tolerations` in newMicroservice

pkg/components/microservice.go
qurname2: fixed

pkg/components/microservice.go
qurname2: fixed

l0kix2 (approved):

Issue #307, opened 2024-07-03
Set observedGeneration in ytsaurus resource status


l0kix2:
To distinguish if the operator already has updated all resources and resource is `Running and happy` or not yet have received new update.

l0kix2:
Common idea can be read on the internet, for example:
https://alenkacz.medium.com/kubernetes-operator-best-practices-implementing-observedgeneration-250728868792

To test this we could check that observedGeneration is expected in the [simple update scenario e2e](https://github.com/ytsaurus/ytsaurus-k8s-operator/blob/main/test/e2e/ytsaurus_controller_test.go#L299-L301)

PR #333, opened 2024-09-06
Add observedGeneration field to the YtsaurusStatus

wilwell:
To distinguish better between states of YtSaurus while updating we are adding observedGeneration to the YtsurusStatus, so if the Ytsaurus configuration is updated we could easily see it by increased  observedGeneration.

This change resolves the issue #307.

### DIFF
--- a/api/v1/ytsaurus_types.go
+++ b/api/v1/ytsaurus_types.go
@@ -734,8 +734,9 @@ type UpdateStatus struct {
 // YtsaurusStatus defines the observed state of Ytsaurus
 type YtsaurusStatus struct {
 	//+kubebuilder:default:=Created
-	State      ClusterState       `json:"state,omitempty"`
-	Conditions []metav1.Condition `json:"conditions,omitempty"`
+	State              ClusterState       `json:"state,omitempty"`
+	Conditions         []metav1.Condition `json:"conditions,omitempty"`
+	ObservedGeneration int64              `json:"observedGeneration,omitempty"`
 
 	UpdateStatus UpdateStatus `json:"updateStatus,omitempty"`
 }
--- a/controllers/sync.go
+++ b/controllers/sync.go
@@ -529,7 +529,11 @@ func (r *YtsaurusReconciler) Sync(ctx context.Context, resource *ytv1.Ytsaurus) 
 		if !componentManager.needSync() {
 			logger.Info("Ytsaurus has synced and is running now")
 			err := ytsaurus.SaveClusterState(ctx, ytv1.ClusterStateRunning)
-			return ctrl.Result{Requeue: true}, err
+			if err != nil {
+				return ctrl.Result{}, err
+			}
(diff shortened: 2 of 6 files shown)


api/v1/ytsaurus_types.go
```diff
@@ -734,8 +734,9 @@ type UpdateStatus struct {
 // YtsaurusStatus defines the observed state of Ytsaurus
 type YtsaurusStatus struct {
 	//+kubebuilder:default:=Created
-	State      ClusterState       `json:"state,omitempty"`
-	Conditions []metav1.Condition `json:"conditions,omitempty"`
+	State              ClusterState       `json:"state,omitempty"`
+	Conditions         []metav1.Condition `json:"conditions,omitempty"`
+	ObservedGeneration int64              `json:"observedGeneration,omitempty"`
```
koct9i, line 739: Probably it needs '//+optional' annotation here
I'm not shure that "omitempty" provides the same effects for code generation.

koct9i: Nevermind. New new required fields in scheme - should be fine as is.

pkg/apiproxy/ytsaurus.go
```diff
@@ -115,6 +115,17 @@ func (c *Ytsaurus) SaveClusterState(ctx context.Context, clusterState ytv1.Clust
 	return nil
 }
 
+func (c *Ytsaurus) SaveObserverGeneration(ctx context.Context, generation int64) error {
+	logger := log.FromContext(ctx)
+	c.ytsaurus.Status.ObservedGeneration = generation
+	if err := c.apiProxy.UpdateStatus(ctx); err != nil {
+		logger.Error(err, "unable to update Ytsaurus cluster status")
```
l0kix2, line 122: I think we should either log error or return, otherwise we have logged it twice.
If you want to add extra context, I suggest to wrap the error with `return fmt.Errorf("failed to set status: %w", err)`

controllers/sync.go
```diff
@@ -529,7 +529,11 @@ func (r *YtsaurusReconciler) Sync(ctx context.Context, resource *ytv1.Ytsaurus)
 		if !componentManager.needSync() {
 			logger.Info("Ytsaurus has synced and is running now")
 			err := ytsaurus.SaveClusterState(ctx, ytv1.ClusterStateRunning)
-			return ctrl.Result{Requeue: true}, err
+			if err != nil {
+				return ctrl.Result{}, err
+			}
+			err = ytsaurus.SaveObserverGeneration(ctx, resource.ObjectMeta.Generation)
+			return ctrl.Result{}, err
```
l0kix2, line 536: AFAIK it is a bad practice to request requeue by by the operator, though let's leave `Requeu: true` here as was before, since it is unintended/unrelated to the current issue change in behaviour.

l0kix2 (changes requested):

l0kix2 (approved):

l0kix2:
Fix #307

PR #334, opened 2024-09-09
Fix ytop-chart-nightly publishing

wilwell:
I've added additional variable `OPERATOR_CHART_NAME`, so nightly helm chart will be published with the correct config and other name.


wilwell:
cc @qurname2

l0kix2 (approved):

PR #336, opened 2024-09-09
Init job creates from InstanceSpec image if specified

### LINKED ISSUE #157: Jobs images must be in sync with component they updating
Not sure if it is a problem for all the components, but it is bug for master exit read only job, because if you run master on 23.1 ytsaurus image and master read only job on 23.2 ytsaurus image job hangs because there is no such command in 23.1.

Maybe all nil instance spec images must be set to core in default webhook and we can use instanceSpec.Image in job and server.

wilwell:
Before this change if resource(master/chyt/scheduler etc.) is updated the init job still uses outdated image from `Spec.CoreImage`. 

This change corrects this behavior, and if the more relevant image exists it gonna be used instead (relevant image for each resource)

Closes  #157

### DIFF
--- a/pkg/components/helpers.go
+++ b/pkg/components/helpers.go
@@ -227,3 +227,10 @@ func AddInitContainersToPodSpec(initContainers []string, podSpec *corev1.PodSpec
 	podSpec.InitContainers = append(containers, podSpec.InitContainers...)
 	return nil
 }
+
+func SelectTrueImage(coreImage string, instanceImage *string) string {
+	if instanceImage != nil {
+		return *instanceImage
+	}
+	return coreImage
+}
--- a/pkg/components/helpers.go
+++ b/pkg/components/helpers.go
@@ -228,9 +228,9 @@ func AddInitContainersToPodSpec(initContainers []string, podSpec *corev1.PodSpec
 	return nil
 }
 
-func SelectTrueImage(coreImage string, instanceImage *string) string {
-	if instanceImage != nil {
-		return *instanceImage
+func getImageWithDefault(componentImage *string, defaultImage string) string {
+	if componentImage != nil {
+		return *componentImage
 	}
-	return coreImage
+	return defaultImage
 }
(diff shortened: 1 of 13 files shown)


pkg/components/init_job.go
```diff
@@ -56,7 +56,12 @@ func NewInitJob(
 	conditionsManager apiproxy.ConditionManager,
 	imagePullSecrets []corev1.LocalObjectReference,
 	name, configFileName, image string,
+	instanceImage *string,
```
l0kix2, line 59: I think there is a wrong place/responsibility of the job object to know about two possible images and decide which one to use. Let's stick with one image argument in the job and pass the correct one from the outside (it will require some copy-paste or helper method, because it is the same logic in different components)

l0kix2 (changes requested):

pkg/components/init_job.go
wilwell: Done

pkg/components/helpers.go
```diff
@@ -227,3 +227,10 @@ func AddInitContainersToPodSpec(initContainers []string, podSpec *corev1.PodSpec
 	podSpec.InitContainers = append(containers, podSpec.InitContainers...)
 	return nil
 }
+
+func SelectTrueImage(coreImage string, instanceImage *string) string {
```
l0kix2, line 231: suppose it can be non-exported since we are using it in the components package only

l0kix2: And the name TrueImage sounds like some soft for windows. Let's maybe call It `SelectEffectiveImage`.
Also, I would expect a signature like `getImageWithDefault(componentImage, defaultImage string)`

l0kix2 (changes requested):

l0kix2 (approved):

l0kix2:
I've realized that there is an issue with this change since all jobs require yt cli installed, but strawberry and UI have different images which can't be used for k8s job. I'll fix that

PR #346, opened 2024-09-16
Fix observed generation test

wilwell:
The test should look on observed generation, not  object generation.

cc @l0kix2

### DIFF
--- a/test/e2e/ytsaurus_controller_test.go
+++ b/test/e2e/ytsaurus_controller_test.go
@@ -1097,7 +1097,7 @@ func getSimpleUpdateScenario(namespace, newImage string) func(ctx context.Contex
 		DeferCleanup(deleteYtsaurus, ytsaurus)
 		name := types.NamespacedName{Name: ytsaurus.GetName(), Namespace: namespace}
 		deployAndCheck(ytsaurus, namespace)
-		oldGeneration := ytsaurus.ObjectMeta.Generation
+		oldGeneration := ytsaurus.Status.ObservedGeneration
 
 		By("Run cluster update")
 		podsBeforeUpdate := getComponentPods(ctx, namespace)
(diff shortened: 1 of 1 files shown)


test/e2e/ytsaurus_controller_test.go
```diff
@@ -1117,10 +1117,14 @@ func getSimpleUpdateScenario(namespace, newImage string) func(ctx context.Contex
 
 		podsAfterFullUpdate := getComponentPods(ctx, namespace)
 		podDiff := diffPodsCreation(podsBeforeUpdate, podsAfterFullUpdate)
+		newYt := ytv1.Ytsaurus{}
+		_ = k8sClient.Get(ctx, name, &newYt)
```
l0kix2, line 1121: Please add check that there is no error here.

test/e2e/ytsaurus_controller_test.go
```diff
@@ -1097,7 +1097,7 @@ func getSimpleUpdateScenario(namespace, newImage string) func(ctx context.Contex
 		DeferCleanup(deleteYtsaurus, ytsaurus)
 		name := types.NamespacedName{Name: ytsaurus.GetName(), Namespace: namespace}
 		deployAndCheck(ytsaurus, namespace)
-		oldGeneration := ytsaurus.ObjectMeta.Generation
+		oldGeneration := ytsaurus.Status.ObservedGeneration
```
l0kix2, line 1100: I think we better test that `ytsaurus.ObjectMeta.Generation` (not ObservedGeneration) of object we have after applying spec for update (I think `k8sClient.Update(ctx, ytsaurus)` should set new value for Generation field in the struct) is the same that we have in `ytsaurus.Status.ObservedGeneration` after the update is finished. This is the scenario we plan to use that field.

l0kix2 (changes requested):

test/e2e/ytsaurus_controller_test.go
wilwell: Done

l0kix2 (approved):

PR #347, opened 2024-09-19
Separate CHYT init options into makeDefault and createPublicClique

achulkov2:
Currently, the `makeDefault` parameter in the CHYT spec has two meanings. First, it [adds](https://github.com/ytsaurus/ytsaurus-k8s-operator/blob/54143b9ad022f0a2068107be8be7d8f9846db607/pkg/components/chyt.go#L106) `--make-default` to the init script parameters, which [causes](https://github.com/ytsaurus/ytsaurus/blob/main/yt/docker/ytsaurus/setup_cluster_for_chyt.sh#L53) the uploaded image to be marked as default for cliques. Second, it [controls](https://github.com/ytsaurus/ytsaurus-k8s-operator/blob/54143b9ad022f0a2068107be8be7d8f9846db607/pkg/components/chyt.go#L191) whether `initChPublicJob` is run. These effects are completely unrelated and one does not always want both of them. This PR moves the second effect under its own option `createPublicClique`.


achulkov2:
I did not rename `makeDefault` for better compatibility with the current setup. I do agree that its name is rather ambiguous as well, maybe we should rename it (now or sometime later).

api/v1/chyt_types.go
```diff
@@ -40,8 +40,12 @@ type ChytSpec struct {
 
 	Ytsaurus *corev1.LocalObjectReference `json:"ytsaurus,omitempty"`
 	Image    string                       `json:"image,omitempty"`
+	// Mark specified image as default for cliques.
 	//+kubebuilder:default:=false
 	MakeDefault bool `json:"makeDefault"`
+	// Create ch_public clique, which is used by default when running CHYT queries.
+	//+kubebuilder:default:=false
+	CreatePublicClique bool `json:"createPublicClique"`
```
l0kix2, line 48: I suggest we make it `*bool` with `,omitempty` tag, so it would be possible to serialize spec to yaml/json and distinguish not set value from true/false. 
And I think the default value better be controlled on the code level, not spec level. That way we can change it if we want for users, who didn't set it, and if we have simple bool — all users will have false in their resource after the operator update.

I know that most other booleans in our spec don't work that way, but I think `*bool` is more flexible here, so new fields better be added that way.

achulkov2: Sure.

l0kix2 (approved):

Issue #349, opened 2024-10-01
Add strawberry apps configuration


l0kix2:
Let's start with something generic addition to the StrawberrySpec which would correctly configure strawberry itself and create k8s svc.

l0kix2:
@dmi-feo maybe this one also somehow related to strawberry configuration and may be fixed https://github.com/ytsaurus/ytsaurus-k8s-operator/issues/42 with you future changeset

PR #355, opened 2024-10-09
Support different controller families in strawberry configuration

dmi-feo:
https://github.com/ytsaurus/ytsaurus-k8s-operator/issues/349


pkg/consts/defaults.go
```diff
@@ -23,3 +23,9 @@ const DefaultName = "default"
 const DefaultMedium = "default"
 
 const MaxSlotLocationReserve = 10 << 30 // 10GiB
+
+const DefaultStrawberryControllerFamily = "chyt"
+
+func GetDefaultStrawberryControllerFamilies() []string {
+	return []string{"chyt", "jupyt"}
+}
```
l0kix2, line 31: I don't understand, should the default family be `chyt` or list of of `chyt, jupyt`?
Maybe first one better be called just `DefaultStrawberryFamily`, not ControllerFamily according to static config values?

l0kix2 (commented):
LGTM, though I want someone else who knows a bit about strawberry look at this.
@psushin does this change helps for you with livy?

pkg/consts/defaults.go
dmi-feo: Poor naming :(
I renamed DefaultStrawberryControllerFamily to DefaultRouteStrawberryControllerFamily - that's explains better what it is about (and it is about the family that will receive requests for domains that are not explicitly specified in `http_controller_mappings`)

api/v1/ytsaurus_types.go
```diff
@@ -517,10 +517,13 @@ type QueryTrackerSpec struct {
...
-	Resources    corev1.ResourceRequirements `json:"resources,omitempty"`
-	Image        *string                     `json:"image,omitempty"`
-	Tolerations  []corev1.Toleration         `json:"tolerations,omitempty"`
-	NodeSelector map[string]string           `json:"nodeSelector,omitempty"`
+	Resources          corev1.ResourceRequirements `json:"resources,omitempty"`
+	Image              *string                     `json:"image,omitempty"`
+	Tolerations        []corev1.Toleration         `json:"tolerations,omitempty"`
+	NodeSelector       map[string]string           `json:"nodeSelector,omitempty"`
+	ExternalProxy      *string                     `json:"externalProxy,omitempty"`
+	ControllerFamilies []string                    `json:"controllerFamilies,omitempty"`
```
koct9i, line 525: Please add comment with examples for this field and next one - comment goes into CRD description and documentation. I.e. `// Supported strawberry families, for example: "chyt", "foo", "bar"`.

l0kix2:
Let's add docstrings and merge it i guess?
@dmi-feo

l0kix2 (approved):

PR #373, opened 2024-10-25
Fix observed generation

koct9i:
- Fix updates of ObservedGeneration
  ObservedGeneration must show observed generation, not reached one.
  I.e. we must update it as soon as spec update is reflected in status.
  
- tests: cleanup tracking of pods changes
  After update wait for observed generation and check state after that.
  This allows to catch wrong state instantly without waiting timeout.


controllers/sync.go
```diff
@@ -602,21 +604,13 @@ func (r *YtsaurusReconciler) Sync(ctx context.Context, resource *ytv1.Ytsaurus)
 
 		logger.Info("Ytsaurus update was finished and Ytsaurus is running now")
 		err := ytsaurus.SaveClusterState(ctx, ytv1.ClusterStateRunning)
-		if err != nil {
-			return ctrl.Result{}, err
-		}
-		err = ytsaurus.SaveObserverGeneration(ctx, resource.ObjectMeta.Generation)
-		return ctrl.Result{}, err
+		return ctrl.Result{Requeue: true}, err
```
wilwell, line 607: Why do we want to requeue here and in `ClusterStateReconfiguration`? If I understand correctly after change of status.  I suppose you just explicitly  set the requeue here.

pkg/apiproxy/ytsaurus.go
```diff
@@ -115,13 +115,12 @@ func (c *Ytsaurus) SaveClusterState(ctx context.Context, clusterState ytv1.Clust
 	return nil
 }
 
-func (c *Ytsaurus) SaveObserverGeneration(ctx context.Context, generation int64) error {
-	c.ytsaurus.Status.ObservedGeneration = generation
-	if err := c.apiProxy.UpdateStatus(ctx); err != nil {
-		return fmt.Errorf("failed to save observed generation in status: %w", err)
+func (c *Ytsaurus) SyncObservedGeneration() bool {
```
wilwell, line 118: Please add doc line to determine whenever we return true or false.

wilwell (changes requested):

controllers/sync.go
koct9i: Spec might already be updated yet another time.
It's easier to requeue than inline all handling at place.

At next iteration operator does final check that we have reached desired state and do final update of observed generation

or starts next update and also does update of observed generation

pkg/apiproxy/ytsaurus.go
koct9i: Good point

wilwell (approved):

wilwell:
@koct9i  please resolve the conflict with `test/e2e/ytsaurus_controller_test.go`, so I can merge the PR

PR #376, opened 2024-10-29
Added namespacedScope value to the helm chart

qurname2:
For use cases when we have several YT clusters within one k8s cluster and would like to deploy one yt-operator per each YT cluster, we have to make ValidatingWebhookConfiguration watch only a namespace where YT and operator are deployed.
If the namespaceScoped variable is true, WATCH_NAMESPACE will be equal to the namespace where the operator is deployed and where Ytsaurus and/or remote resources are placed. The same logic is applied for namespaceSelector in ValidatingWebhookConfiguration.


kruftik (approved):

koct9i (approved):
LGTM

Issue #325, opened 2024-08-19
UpdateSelector needs to be decoupled


Krock21:
UpdateSelector has fixed amount of values now that do not cover all combinations of possible and reasonable updates.

In my case I wanted to update HttpProxy and QueryTrackers, and only UpdateSelector: StatelessOnly allows me to do it and it forces me to update execNodes and DataNodes too.

I suggest either:
- Making UpdateSelector masked and composable from every component, like `QUERY_TRACKER | HTTP_PROXY`
- Splitting it to many different selectors who are `true` by default: `updateQueryTracker: true; updateHttpProxy=true...`

koct9i:
```
type ComponentUpdateSelector struct {
  Component CompontntKind // required enum
  Name string // optional instance group name
  Update true // default false
}

type YtsaurusSpec struct {
  ...
  UpdateSelectors []ComponentUpdateSelector
  ...
}
```

l0kix2:
I guess some groups of component kinds should be supported (something like current StatelessOnly and stuff, but need think bit better about it than last time)

koct9i:
As I see there are three levels:
1. Masters
2. Stateful - DataNodes and TabletNodes
3. Stateless - all others

So, we could rearrange existing enum to reflect that.
And reuse it in new selector.

l0kix2:
Also we need:
1) to preserve current default behaviour: "update everything" for users who don't care about selective updates
2) to have "all updates blocked" switch.

From the config ux/code perspective it would be convienient to have empty value as nothing and group selector "Everything", but it will require "Everything" to be default value here on crd level (which I don't like very much, but guess it is fine). I don't know if list field can have default value though.

koct9i:
Does not update and show explaining status is safe and sane default option.

l0kix2:
> Does not update and show explaining status is safe and sane default option.
I don't know if our users is expecting that they will need to expicitly set field in their spec to be able to update cluster, but maybe it is ok. Need to discuss

koct9i:
I expect that all known users are smart enough.

l0kix2:
I guess if we are adding new field we can switch the default to "No updates" if we introduce some kind of "UpdateBlocked" status for the resource (with event maybe)

l0kix2:
So to sum things up:
[...]

PR #383, opened 2024-11-04
Add multiple update selectors

wilwell:
This PR introduces the functionality of multiple update selectors.

Instead of outdated `UpdateSelector` users can use `UpdatePlan` with list of `ComponentUpdateSelector` inside it.
Each  `ComponentUpdateSelector` can either contain `ComponentClass` or `Component`.  

`ComponentClass` contains `Everything, Nothing, Stateless` groups of component. If operator sees `Nothing` component class, then all updates are forbidden. 
`Component` could has `Type` and  `Name`  fields. If `Name` is not set then all components with this `Type` will be updated.

Also I've added webhook check, that empty `UpdatePlan` is forbidden. In case if `UpdatePlan` is not set operator uses values of deprecated `UpdateSelector` for smoother transition between versions.


controllers/sync.go
```diff
@@ -393,60 +393,75 @@ type updateMeta struct {
...
+		return ytv1.UpdateFlowMaster
 	}
+	if component == consts.TabletNodeType {
+		return ytv1.UpdateFlowTabletNodes
+	}
+	if component == consts.DataNodeType || component == consts.ExecNodeType {
+		return ytv1.UpdateFlowFull
+	}
+	return ytv1.UpdateFlowStateless
+}
```
l0kix2, line 407: I see this PR is raw yet, but just want to emphasize some things.
Obviously to add updateSelectors field is not challenging part of the issue, but to correctly update each component (or combination of components) selected by it, so all required pre and post steps were executed.

Currently we have some flows for component updates:
**master**
- PossibilityCheck
- WaitingForSafeModeEnabled
- WaitingForSnapshots
- WaitingForPodsRemoval
- WaitingForPodsCreation
- WaitingForMasterExitReadOnly
- WaitingForSafeModeDisabled

**tablet nodes**
- PossibilityCheck
- WaitingForTabletCellsSaving
- WaitingForTabletCellsRemovingStart
- WaitingForTabletCellsRemoved
- WaitingForPodsRemoval
- WaitingForPodsCreation
- WaitingForTabletCellsRecovery

**stateless**
- WaitingForPodsRemoval
- WaitingForPodsCreation
- WaitingForOpArchiveUpdatingPrepare
- WaitingForOpArchiveUpdate
- WaitingForQTStateUpdatingPrepare
- WaitingForQTStateUpdate

**everything**
- PossibilityCheck
- WaitingForSafeModeEnabled
- WaitingForTabletCellsSaving
- WaitingForTabletCellsRemovingStart
- WaitingForTabletCellsRemoved
- WaitingForSnapshots
- WaitingForPodsRemoval
- WaitingForPodsCreation
- WaitingForMasterExitReadOnly
- WaitingForTabletCellsRecovery
- WaitingForOpArchiveUpdatingPrepare
- WaitingForOpArchiveUpdate
- WaitingForQTStateUpdatingPrepare
- WaitingForQTStateUpdate
- WaitingForSafeModeDisabled

As you see `everything` includes (basically copy-pastes) parts `master`,`tnds`,`stateless` flows.
And 
if you want implement `UpdateSelectors = {component=scheduler}` you have to add flow for scheduler with `WaitingForOpArchiveUpdatingPrepare/WaitingForOpArchiveUpdate/WaitingForQTStateUpdatingPrepare/WaitingForQTStateUpdate`
if you want to implement `UpdateSelectors = {component=qt}` you have to add flow with `WaitingForQTStateUpdatingPrepare/WaitingForQTStateUpdate`
and so on.
So this predefined flows code is a dead end way if we want to achieve update granularity we want. So we should somehow concatenate workflows for particular components.

And i think the most simplest way here is to do this change with introducing Linear update (https://github.com/ytsaurus/ytsaurus-k8s-operator/issues/115).
I assume the flow in the code should look something like that:
```
# update flow
if discoveryType in toUpdateList: // updatable by selector and require update
	remove/create discovery pods

if httpProxy in updateList:
       remove/create hp pods
// tcp proxy
// rpc proxy
// data nodes
// exec nodes
// master cache

if tabletNodes in updateList:
     - PossibilityCheck
     - WaitingForTabletCellsSaving
     - WaitingForTabletCellsRemovingStart
     - WaitingForTabletCellsRemoved
     - WaitingForPodsRemoval
     - WaitingForPodsCreation
     - WaitingForTabletCellsRecovery

// ui
// ca
// yqla (but soon also will have pre-update jobs)
// sch with its flow
// qt with its flow
// qa with its flow
// strawberry
// master with its flow (updating in the end, because it should never be newer than other components)
```
Initializing flow is the same, but master in the beginning.

So basically I think we need here to:
1) resolve updateSelectors in a list of allowed components, match it with all component's statuses so we have a list of components we plan to update, sort it according to predefined order, choose first one we will update in this reconcillation loop, put it in `UpdateStatus.Components` 
2) check current `UpdateStatus.State` and `UpdateStatus.Conditions` and either call components.sync and requeue or set next State of this component flow and requeue.

I understand it is a bit vague overview, but maybe it helps.

l0kix2:
Please rebase on the main branch since https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/387 is merged — which have couple of new stages in the flows

controllers/sync.go
```diff
@@ -393,103 +393,116 @@ type updateMeta struct {
...
```
l0kix2, line 407: Since we have combination of selectors that wouldn't be possible to cover with existing flow (like like `master+http proxies` will go to everything and will update tablet nodes which wasn't allowed by selector) we have to make changes in flows code.
Also we've discovered that we already have checks like `needSchedulerUpdate` and `needQueryTrackerUpdate` we simplify code a bit and just have one "EverythingFlow" with added `if`s for the master and tablet nodes in the required places. Other flows and flow field itself can be removed with this change — that would make less copypaste and possible errors in the code.

The thing which is bothers me that this flows is not test covered very well and it is rather expensive (in the meaning of test time and code amount) to test all the required cases in e2e.
So I suggest we convert this flows code with switch-cases in a different shape. We introduce some struct like this
```
type flowStep struct {
  name string
  condition func(ytv1.Ytsaurus) bool
  onSuccess func() err
}
```
and have some function to build flow from the list of exact components which require update
```go
func buildFlow(components) []flowStep{
    var steps []flowStep
    if components.Contain(masterType) {
        steps = append(flowStep{
             name: "safe mode enabled",
             condition: func(ytsaurus ytv1.Ytsaurus) {
                 return resource.Status.UpdateStatus.State == `ytv1.UpdateStateWaitingForSafeModeEnabled`
             },
             onSuccess: func(ytsaurus ytv1.Ytsaurus) {
                 return ytsaurus.SaveUpdateState(ctx, ytv1.UpdateStateWaitingForTabletCellsSaving)
              }
        })
    }
    ...
    return steps
}
```
(I'm not sure that this exact fields and functions should be there, but something similar should work).

That way we can cover our flow building with fast unittest good enough and cover all the important cases and be sure not to break stuff in flow logic.

So the whole code for this PR will look like this:
1) convert fullUpdate+updateSelector+updateSelectors to the updateSelectors list
2) get components which require update and filter them with updateSelectors — set it to the updating components field of UpdateState and set state=Updating
3) generate flow for the list of the componenents
4) apply the flow

all 1-3 steps can be untitested that way.

I think it is better to do it in the separate PR and after that use that code in this PR.
What do you think? Does it make sense to you?

l0kix2 (changes requested):

controllers/sync.go
wilwell: Done in https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/422

pkg/consts/types.go
```diff
@@ -24,3 +24,14 @@ const (
...
 	SpytType                 ComponentType = "SPYT"
 )
+
+type ComponentGroup string
+
+const (
+	// ComponentGroupStateless group contains only stateless components (not master, data nodes, tablet nodes)
+	ComponentGroupStateless ComponentGroup = "Stateless"
+	// ComponentGroupStateful group contains every component except master
+	ComponentGroupStateful   ComponentGroup = "Stateful"
```
l0kix2, line 34: Statefull can't include stateless, it is confusing.
In the issue @koct9i suggested Stateful to consist of data+tablet nodes https://github.com/ytsaurus/ytsaurus-k8s-operator/issues/325#issuecomment-2429560696
And even that, in my opinion, can be interpreted differently (is it dnds+tnds+master or is it dnds+tnds). 
I suggest not having this group. We never do such update and if this will needed — we write 2 selectors: one for tablet nodes and one for datanodes.

api/v1/ytsaurus_types.go
```diff
@@ -779,15 +782,18 @@ const (
...
 )
 
-type UpdateFlow string
+type ComponentUpdateSelector struct {
+	//+optional
+	ComponentType consts.ComponentType `json:"componentType,omitempty"`
+	//+optional
+	ComponentGroup consts.ComponentGroup `json:"componentGroup,omitempty"`
+	//+optional
+	ComponentName string `json:"componentName,omitempty"`
```
l0kix2, line 791: Maybe if we are in context of components let's just have `Type, Name, Group` fields?

l0kix2, line 791: I have no idea what to put in the Name actually. 
I suggest we add examples here for unnamed component like master, for unnamed datanode and for named datanode.

Also AFAIK we have `Name` which already include type and for sure users will try to put the value which they set in the spec for the respective component. 
Let's think now if we really want "full" names here. If so I guess we should name the field like FullName or smth.

l0kix2 (approved):

api/v1/ytsaurus_types.go
```diff
@@ -672,11 +674,14 @@ type YtsaurusSpec struct {
...
-	//+kubebuilder:validation:Enum={"","Nothing","MasterOnly","DataNodesOnly","TabletNodesOnly","ExecNodesOnly","StatelessOnly","Everything"}
-	// UpdateSelector is an experimental field. Behaviour may change.
-	// If UpdateSelector is not empty EnableFullUpdate is ignored.
+	//+optional
+	// Deprecated: UpdateSelector is going to be removed soon. Please use UpdateSelectors instead.
 	UpdateSelector UpdateSelector `json:"updateSelector"`
 
+	//+optional
+	// Controls the components that should be updated during the update process.
+	UpdatePlan []ComponentUpdateSelector `json:"updateSelectors,omitempty"`
```
koct9i, line 683: Please update json name here

pkg/components/component.go
```diff
@@ -60,14 +61,18 @@ type baseComponent struct {
...
 // For example for master component name is "Master",
 // For data node name looks like "DataNode<NameFromSpec>".
-func (c *baseComponent) GetName() string {
+func (c *baseComponent) GetFullName() string {
 	return c.labeller.GetFullComponentName()
 }
 
+func (c *baseComponent) GetShortName() string {
+	return c.labeller.GetInstanceGroup()
+}
```
l0kix2, line 74: Are you sure that `default` won't be returned here if name not set? Maybe it should be take the code from `getGroupName()` this function to be reused there?

controllers/sync.go
```diff
@@ -15,65 +15,48 @@ import (
...
 	var canUpdate []ytv1.Component
 	var cannotUpdate []ytv1.Component
-	needFullUpdate := false
 
 	for _, comp := range needUpdate {
 		component := ytv1.Component{
-			Name: comp.GetName(),
+			Name: comp.GetShortName(),
 			Type: comp.GetType(),
 		}
```
l0kix2, line 53: Woudln't it be easier if we use ytv1.Component for needUpdate and allComponents and store them that way in ComponentManager, so we convert them in one place?

controllers/sync.go
```diff
@@ -83,29 +66,84 @@ func chooseUpdatingComponents(spec ytv1.YtsaurusSpec, needUpdate []components.Co
 		return nil, "All components are uptodate"
 	}
 
-	switch configuredSelector {
-	case ytv1.UpdateSelectorEverything:
-		if needFullUpdate {
+	if len(configuredSelectors) == 1 && configuredSelectors[0].Class == consts.ComponentClassEverything {
```
l0kix2, line 69: Here we should check for "one of configured selectors is ComponentClassEverything", because we don't reject something like `[{Type: DataNodes, Name: somename}, {Class: Everything}]`.
But I guess all others can be thrown away in `getEffectiveSelectors` if `Everything` is present.

controllers/sync.go
```diff
@@ -83,29 +66,84 @@ func chooseUpdatingComponents(spec ytv1.YtsaurusSpec, needUpdate []components.Co
...
+
+func needFullUpdate(needUpdate []components.Component) bool {
+	statelessSelector := []ytv1.ComponentUpdateSelector{{Class: consts.ComponentClassStateless}}
+	for _, comp := range needUpdate {
+		component := ytv1.Component{
+			Name: comp.GetShortName(),
+			Type: comp.GetType(),
+		}
+		isStateless := canUpdateComponent(statelessSelector, component)
+		if !isStateless && component.Type != consts.DataNodeType {
```
l0kix2, line 88: Data nodes already considered not stateless in canUpdateComponent, so no need to check

api/v1/ytsaurus_types.go
```diff
@@ -811,7 +817,7 @@ type Component struct {
 }
 
 func (c Component) String() string {
-	return c.Name
+	return fmt.Sprintf("%s(%s)", c.Type, c.Name)
```
l0kix2, line 820: This will look like ` updating components: [Master()]` for nameless componentns, let's add name only if not empty

l0kix2:
Looks good overall, have some nitpicks

koct9i (approved):

l0kix2 (approved):

l0kix2:
Fix #325

PR #384, opened 2024-11-04
Prepare for e2e refactoring

koct9i:
- test: build own scheme in e2e
  
- test: regroup e2e tests
  No functional changes, just reordering and adding contexts.
  
- test: retry update in e2e

### DIFF
--- a/test/e2e/suite_test.go
+++ b/test/e2e/suite_test.go
@@ -30,9 +30,10 @@ import (
 	. "github.com/onsi/gomega"
 	otypes "github.com/onsi/gomega/types"
 
-	"k8s.io/client-go/kubernetes/scheme"
-
+	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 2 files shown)


test/e2e/suite_test.go
```diff
@@ -92,12 +94,15 @@ var _ = SynchronizedBeforeSuite(func(ctx context.Context) []byte {
 	Expect(err).NotTo(HaveOccurred())
 	Expect(cfg.Host).To(Equal(string(host)))
 
-	err = ytv1.AddToScheme(scheme.Scheme)
+	scheme := runtime.NewScheme()
```
l0kix2, line 97: Could you please explain how this scheme is different from one registered in scheme package?

koct9i: Do you mean default global scheme?
I'm not sure that it is thread safe.
It's cleaner to build fresh scheme here.

l0kix2 (approved):

PR #386, opened 2024-11-05
Refactor components interfaces

koct9i:
Split layers slightly more clearly

### DIFF
--- a/controllers/component_manager.go
+++ b/controllers/component_manager.go
@@ -34,23 +34,23 @@ func NewComponentManager(
 	ytsaurus *apiProxy.Ytsaurus,
 ) (*ComponentManager, error) {
 	logger := log.FromContext(ctx)
-	resource := ytsaurus.GetResource()
+	resource := ytsaurus.Resource()
 
-	clusterDomain := getClusterDomain(ytsaurus.APIProxy().Client())
+	clusterDomain := getClusterDomain(ytsaurus.Client())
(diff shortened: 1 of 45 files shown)


pkg/components/exec_node.go
```diff
@@ -14,21 +14,22 @@ import (
 )
 
 type ExecNode struct {
+	localServerComponent
 	baseExecNode
-	localComponent
 	master Component
 }
```
l0kix2, line 20: Do you think it is ok to have server twice: inside localServerComponent struct and inside baseExecNode here?

l0kix2 (commented):
Could you please write some motivation behind this change when it would be ready for review (not draft)?

controllers/component_manager.go
```diff
@@ -152,22 +152,23 @@ func NewComponentManager(
 			return nil, err
 		}
 
-		componentStatus, err := c.Status(ctx)
+		componentStatus, err := c.Sync(ctx, true)
```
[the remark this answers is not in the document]
koct9i: This PR in intermediate state so there might be rough pieces. I don't like this one too.
Unfortunately there is no defined clear semantics. I'm trying to make one.

pkg/components/exec_node.go
koct9i: There is no problems with that since this is reference, not instance itself.

PR #387, opened 2024-11-06
Add yqla update job

Krisha11:


api/v1/ytsaurus_types.go
```diff
@@ -696,6 +696,8 @@ const (
 	UpdateStateWaitingForOpArchiveUpdate          UpdateState = "WaitingForOpArchiveUpdate"
 	UpdateStateWaitingForQTStateUpdatingPrepare   UpdateState = "WaitingForQTStateUpdatingPrepare"
 	UpdateStateWaitingForQTStateUpdate            UpdateState = "WaitingForQTStateUpdate"
+	UpdateStateWaitingForYqlAUpdatingPrepare      UpdateState = "WaitingForYqlAUpdatingPrepare"
```
l0kix2, line 699: Let's maybe either choose 
`UpdateStateWaitingForYqlAgentUpdatingPrepare` 
or
`UpdateStateWaitingForYqlaUpdatingPrepare`

`...YqlA...` is rather odd abbreviation

Krisha11: Changed to `Yqla`

l0kix2 (approved):

pkg/components/yql_agent.go
```diff
@@ -73,6 +75,18 @@ func NewYQLAgent(cfgen *ytconfig.Generator, ytsaurus *apiproxy.Ytsaurus, master
...
 			getNodeSelectorWithDefault(resource.Spec.YQLAgents.NodeSelector, resource.Spec.NodeSelector),
 		),
+		updateEnvironment: NewInitJob(
+			&l,
+			ytsaurus.APIProxy(),
+			ytsaurus,
+			resource.Spec.ImagePullSecrets,
+			"yql-agent-update-environment",
+			consts.ClientConfigFileName,
+			getImageWithDefault(resource.Spec.YQLAgents.Image, resource.Spec.CoreImage),
```
Krock21, line 85: Taking CoreImage as QT/Yqla image is not good. They have initial version there and it will only lead to problems (and we had them already)

Do we take `CoreImage` as default to QueryTracker too?

If yes, then I am fine with proceeding and creating a separate issue for it.

Krock21 (approved):

pkg/components/yql_agent.go
Krisha11: Yes, we have it everywhere
Like this in init jobs:
https://github.com/ytsaurus/ytsaurus-k8s-operator/blob/main/pkg/components/query_tracker.go#L82

And as a component image too
https://github.com/ytsaurus/ytsaurus-k8s-operator/blob/main/pkg/components/server.go#L96

l0kix2: Since we are only using yt cli in job it should be fine.

PR #388, opened 2024-11-12
kata-compat: mount TLS-related files to a separate directory

kruftik:
current scheme with mounting multiple volumes into /config is not compatible with kata-containers


koct9i:
LGTM

I would prefer to move all this stuff into `/etc/ytsaurus`, dumping into root is just ugly.

koct9i (approved):

PR #392, opened 2024-11-14
add podAntiAffinity for masters in demo cfg, fix yqlAgent count for 0.9.1

futujaos:


config/samples/0.9.1/cluster_v1_demo.yaml
```diff
@@ -15,6 +15,16 @@ spec:
   primaryMasters:
     instanceCount: 3
     cellTag: 1
+    affinity:
```
futujaos, line 18: without it, this error will appear:

```
The Ytsaurus "ytdemo" is invalid: spec.primaryMasters.affinity.podAntiAffinity: Required value: Masters should be placed on different nodes
```

i added fix to 0.9.1 and to file without version (is it latest version?), don't know if it's applicable to previous version

l0kix2:
Oh, hi, Oleg

l0kix2:
@koct9i what do you think? Maybe `ephemeralCluster` should be set in this sample instead of antiaffinity if it is just for demo purposes?

l0kix2:
Not sure if we should change 0.9.1 also, i guess it must be checked with 0.9.1 operator and maybe it doesn't have such check yet. Not sure if we even need samples for different versions (especially old ones), I guess we need to preserve latest and maybe some of the previous and update them on release (it would also be good to test them automatically).

futujaos:
> Maybe `ephemeralCluster` should be set in this sample instead of antiaffinity if it is just for demo purposes?

it's up to you (that is `ephemeralCluster`?), i just think there should be an example with multiple masters somewhere in docs, currently all examples contains only single master

koct9i:
Probably we could resurrect default anti-affinity for masters (but not in form defaulter webhook) and adjust validation.

PR #395, opened 2024-11-19
Use expected instance count as default for minimal ready count

koct9i:
Currently operator has silly assumption that all pods are always created.

### DIFF
--- a/controllers/ytsaurus_local_test.go
+++ b/controllers/ytsaurus_local_test.go
@@ -6,6 +6,8 @@ import (
 	"testing"
 	"time"
 
+	"k8s.io/utils/ptr"
+
 	"github.com/stretchr/testify/require"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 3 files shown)


l0kix2:
Yes, last time i tried to fix it some weird case (from tests i guess) appear which was covered by this odd code. Can't remember now. But I guess it shouldn't break things

controllers/ytsaurus_local_test.go
```diff
@@ -56,6 +58,11 @@ func TestYtsaurusFromScratch(t *testing.T) {
 	defer h.Stop()
 
 	ytsaurusResource := testutil.BuildMinimalYtsaurus(namespace, ytsaurusName)
+	// FIXME(khlebnikov): This test is broken by design.
```
l0kix2, line 61: What happens if min ready not set to zero? It never becomes ready? What prevents test to work with default nil value?

koct9i: This test runs without controller bundle - without statefulset controller nobody creates pods.

l0kix2: ah, yes, no point of waiting running state here

l0kix2 (approved):

PR #401, opened 2024-12-13
Fix updateSelector comp type

l0kix2:


Krock21:
A test for a cluster with custom names for all components would be nice. With various cluster updates

PR #405, opened 2024-12-18
Do not require specifying quota when specifying location watermarks

p4vook:


pkg/ytconfig/node.go
```diff
@@ -325,7 +325,8 @@ func getDataNodeServerCarcass(spec *ytv1.DataNodesSpec) (DataNodeServer, error)
 		}
 		if quota != nil {
 			storeLocation.Quota = *quota
-
+		}
+		if location.LowWatermark != nil or storeLocation.Quota != nil {
```
jasminetealover, line 329: Let's also move `storeLocation.MaxTrashTtl = location.MaxTrashMilliseconds` out of this `if`.

jasminetealover (approved):

p4vook:
@l0kix2 could you please check this PR out? It still says that 1 approval is required.

l0kix2:
> @l0kix2 could you please check this PR out? It still says that 1 approval is required.

sure, sorry for the delay, will look into it

pkg/ytconfig/node.go
```diff
@@ -337,8 +338,8 @@ func getDataNodeServerCarcass(spec *ytv1.DataNodesSpec) (DataNodeServer, error)
 			storeLocation.HighWatermark = storeLocation.LowWatermark / 2
 			storeLocation.DisableWritesWatermark = storeLocation.HighWatermark / 2
 			storeLocation.TrashCleanupWatermark = storeLocation.LowWatermark
-			storeLocation.MaxTrashTtl = location.MaxTrashMilliseconds
 		}
```
l0kix2, line 341: It looks a bit hard to understand and error-prone with all these nested ifs.
I suggest we rewrite it in less nested way:
```go
// setting quota
if quota != nil {
    storeLocation.Quota = *quota
}

// setting low watermark
if location.LowWatermark != nil {
    storeLocation.LowWatermark = location.LowWatermark.Value()
} else if quota != nil {
    gb := float64(1024 * 1024 * 1024)
    storeLocation.LowWatermark = int64(math.Min(0.1*float64(storeLocation.Quota), float64(25)*gb))
}

// setting other watermarks if low watermark is set
if storeLocation.LowWatermark > 0 {
    storeLocation.HighWatermark = storeLocation.LowWatermark / 2
    storeLocation.DisableWritesWatermark = storeLocation.HighWatermark / 2
    storeLocation.TrashCleanupWatermark = storeLocation.LowWatermark
}


```

l0kix2: Also I suggest to add test cases for scenarios which aren't currently covered, so this  code won't become broken in the next changes.
Tests for config generator are rather easy to write: just add some test cases with specs having quota=nil + low_watermark=nil, quota!=nil + low_watermark=nil, quota=nil + low_watermark!=nil 
and simply check that canonized result looks as expected in config.

l0kix2 (changes requested):

pkg/ytconfig/node.go
p4vook: Sorry, just got to it. Where should I put the test?

l0kix2: no worries, here are the config tests https://github.com/ytsaurus/ytsaurus-k8s-operator/blob/main/pkg/ytconfig/generator_test.go

PR #407, opened 2024-12-26
Support the ability to deploy a Kafka proxy

savnadya:


pkg/consts/address.go
```diff
@@ -60,4 +61,8 @@ const (
 
 	MasterCachesRPCPort        = 9018
 	MasterCachesMonitoringPort = 10018
+
+	KafkaProxyMonitoringPort = 10020
```
zlobober, line 65: Two requests here.
1. Monitoring port should be equal to YT RPC port + 1000.
2. I am almost sure that 9020, 9021 and 9023 are occupied by some other services (maybe through Y internal configuration system). AFAIR, 9020 is a standard port for the discovery servers, therefore there will be a clash of monitoring ports. I'd prefer not to introduce collisions or inconsistencies with the internal Y conventions here, so I suggest choosing higher two ports (rpc and kafka), ideally the same as you already have inside.

l0kix2 (approved):

l0kix2:
If @zlobober is happy with the ports we end up with, let's merge

PR #409, opened 2024-12-28
Configure yqla mrjob syslibs

Krisha11:
This code is incompatible with old QueryTracker images. Before this we have not supported this options in yqla config.

Also current nightly QueryTracker images are broken without this changes. It is the easiest way to fix them.

That means that new k8s release will only be compatible with new QT release. And this should be specified in release notes.

### DIFF
--- a/api/v1/ytsaurus_types.go
+++ b/api/v1/ytsaurus_types.go
@@ -538,6 +538,11 @@ type StrawberryControllerSpec struct {
 
 type YQLAgentSpec struct {
 	InstanceSpec `json:",inline"`
+
+	// Sets mrjob dynamic library paths in yql agent config.
+	//+kubebuilder:default:=false
+	//+optional
+	ConfigureMrJobDynamicLibraries bool `json:"configureMrJobDynamicLibraries"`
 }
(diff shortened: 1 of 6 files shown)


l0kix2:
Current change would force users to update yqla image with ytop update, but it may be not what they want if the old image works fine. 
I suggest we add some flag in yqla spec, which would result in adding or not adding new fields in yqla config.
We may go either with:
  1) add new field by default and give users the option to disable that in the spec — that way users, who haven't read release notes carefully might update ytop and still break yqla
  2) don't add new field by default and give users the option to enable that in the spec — that way users can safely update ytop, but when they update yqla they should read yqla image release notes and enable the flag (and in the future when we expect most of the users migrate we can switch default to the empty flag).
I'm fine with both options, since you're the most involved in supporting opensource yqla users, but I think we should implement one them and not block ytop update on yqla image update.

Krisha11:
> I'm fine with both options, since you're the most involved in supporting opensource yqla users, but I think we should implement one them and not block ytop update on yqla image update.

Implemented the second one

PR #412, opened 2025-01-08
Enable real chunks job

### LINKED ISSUE #396: Enable enable_real_chunk_locations before update to 24.2
Prior to datanodes update we should add a k8s job, that would set `//sys/@config/node_tracker/enable_real_chunk_locations=%true` if image version is in a range `23.2 — 24.1`. It is required to enable update to 24.2 in the future. 

After that we will ask users to update ytop to the version with this fix and run update (full or data nodes only) @ 23.2 or 24.1 version.

N.B. Such update should be done with re-creating all ytsaurus nodes pods, which is default behaviour for full update (which is triggered when master image is updated). Node restart wouldn't happen with `updateSelector=MasterOnly`, but we consider `updateSelector` experimental functionality for those who knows what they doing (for example they can restart nodes themselves to have less downtime).

Extra details:
-  job should run on both cluster create (Initialization state) and update (Updating state) before dnd pods are created/recreated
- remote data nodes also should have the same workflow (though if they exist they will require separate update with new ytop version)

l0kix2:
Fix #396

I've added k8s job in the master component. The job simply checks if ytsaurus image version is between 23.2—24.1 including and if yes — sets `//sys/@config/node_tracker/enable_real_chunk_locations=%true` according to the issue.
It is being executed:
 - when the cluster is `Initilazing` (after master pods are created and before dnd pods are created)
 - on each master update  (in FullUpdate mode since we want dnds to be restarted also) after master in safe mode but before all pods are recreated (because we want flag before dnd restart).

This is required to prepare user clusters for 24.2. We will ask users to update their cluster to some (23.2—24.1) version with ytop version with this code prior to 24.2 update.

### How it was tested
  1. I've added that flag is set to true after update for current versions we test against (create 23.2.0 and update to 23.2.1 + create 23.2.0 and update to 24.1.0) + no node with `use_imaginary_chunk_locations=true` exist after the update.
  2. I've manually tested that flag is not set for 23.1 (we don't test against it, kinda unsuported i'd say)
  3. I've manually tested that update from 23.1 (flag false) to 23.2 sets flag to true.

### Other
We also may implement some check for possibility of update to 24.2 checking data nodes meta information in a separate issue.


achulkov2 (approved):
All contents LGTM, but I think it is useful to add a validation job that would stop the update to 24.2 if it would lead to master cores.

PR #416, opened 2025-01-09
Support not creating non-existing users

l0kix2:
Supporting overriding default behaviour for OAuth user creation implemented in https://github.com/ytsaurus/ytsaurus/pull/997


pkg/ytconfig/generator.go
```diff
@@ -670,10 +677,19 @@ func (g *Generator) getHTTPProxyConfigImpl(spec *ytv1.HTTPProxiesSpec) (HTTPProx
...
+		var createUserIfNotExist *bool
+		if oauthService.DisableUserCreation != nil {
+			createUserIfNotExist = ptr.To(!*oauthService.DisableUserCreation)
+		}
+		c.Auth.OauthCookieAuthenticator = &OauthCookieAuthenticator{
+			CreateUserIfNotExists: createUserIfNotExist,
+		}
+		c.Auth.OauthTokenAuthenticator = &OauthTokenAuthenticator{
+			CreateUserIfNotExists: createUserIfNotExist,
+		}
```
Krock21, line 692: Am I right that if `oauthService.DisableUserCreation == nil`, then `createUserIfNotExist` will stay `nil`, and `c.Auth.OauthCookieAuthenticator.CreateUserIfNotExists` will stay nil, and in resulting spec `create_user_if_not_exists` will not appear

pkg/ytconfig/canondata/TestGetHTTPProxyConfig/test.canondata
```diff
@@ -109,8 +109,10 @@
             ];
         };
         "oauth_cookie_authenticator"={
+            "create_user_if_not_exists"=%false;
```
Krock21, line 112: What is this canondata? Does it mean this is default behaviour if a user did not specify `DisableUserCreation`?

`create_user_if_not_exists` should be not set if user did not specify `DisableUserCreation`, and set to `!DisableUserCreation` otherwise

Krock21 (commented):
Tests for it would be nice. Something like "Verify that `DisableUserCreation == nil/true/false` leads to `create_user_if_not_exists = nil/false/true` in http proxy spec

pkg/ytconfig/generator.go
l0kix2: yes, that's the idea, on nil we just proxying default server behavior without making assumptions for default or forcing something.

pkg/ytconfig/canondata/TestGetHTTPProxyConfig/test.canondata
l0kix2: no, i've added true in test oauth spec

l0kix2:
> Tests for it would be nice. Something like "Verify that `DisableUserCreation == nil/true/false` leads to `create_user_if_not_exists = nil/false/true` in http proxy spec

I've removed setting the new field in common test OAuth spec (so it is nil) — so there is no diff for existing test, but added two extra test with true/false values and respective false/true in resulting canondata.

config/crd/bases/cluster.ytsaurus.tech_ytsaurus.yaml
```diff
@@ -14893,6 +14893,10 @@ spec:
                 type: object
               oauthService:
                 properties:
+                  disableUserCreation:
+                    description: If DisableUserCreation is set, proxies will NOT create
+                      non-existing users with O
```
Krock21, line 14898:
```suggestion
                      non-existing users with OAuth authentication
```

Krock21 (approved):

config/crd/bases/cluster.ytsaurus.tech_ytsaurus.yaml
l0kix2: weird, will check, though it is auto-generated from comment string, I can't fix it here, it will be diff next time

l0kix2: https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/418

PR #420, opened 2025-01-10
Added DNSConfig into Instance and YTsaurusSpec

qurname2:


api/v1/ytsaurus_types.go
```diff
@@ -659,8 +662,10 @@ type YtsaurusSpec struct {
 
 	NodeSelector map[string]string   `json:"nodeSelector,omitempty"`
 	Tolerations  []corev1.Toleration `json:"tolerations,omitempty"`
-
-	Bootstrap *BootstrapSpec `json:"bootstrap,omitempty"`
+	// DNSConfig allows customizing the DNS settings for the pods.
+	//+optional
+	DNSConfig *corev1.PodDNSConfig `json:"dnsConfig,omitempty"`
```
l0kix2, line 667: This field looks unused, I suppose it should be used as a fallback

pkg/components/server.go
```diff
@@ -349,8 +349,11 @@ func (s *serverImpl) rebuildStatefulSet() *appsv1.StatefulSet {
 		Affinity:     s.instanceSpec.Affinity,
 		NodeSelector: s.instanceSpec.NodeSelector,
 		Tolerations:  s.instanceSpec.Tolerations,
+		DNSConfig:    s.instanceSpec.DNSConfig,
```
l0kix2, line 352: Hm, we don't have fallback here to main resource value. Though we don't have for NodeSelector, Toleration, etc either. Will need to add that later, I don't think it should be duplicated in all the components.

pkg/components/server.go
```diff
@@ -349,8 +349,11 @@ func (s *serverImpl) rebuildStatefulSet() *appsv1.StatefulSet {
...
 	}
-	if ptr.Deref(s.instanceSpec.HostNetwork, s.commonSpec.HostNetwork) {
+	if s.instanceSpec.DNSPolicy != nil {
+		statefulSet.Spec.Template.Spec.DNSPolicy = *s.instanceSpec.DNSPolicy
+	} else if ptr.Deref(s.instanceSpec.HostNetwork, s.commonSpec.HostNetwork) {
```
l0kix2, line 356: Why we don't set hostnetwork if DNSPolicy is not set? I don't see a connection.

pkg/components/server.go
qurname2: Yes, but I would suggest to do that in different PR for everything at once - NodeSelector, Tolerationd, DNSConfig

api/v1/ytsaurus_types.go
```diff
@@ -285,6 +285,10 @@ type InstanceSpec struct {
 	// Component config for native RPC bus transport.
 	//+optional
 	NativeTransport *RPCTransportSpec `json:"nativeTransport,omitempty"`
+	// DNSConfig allows customizing the DNS settings for the pods.
+	//+optional
+	DNSConfig *corev1.PodDNSConfig `json:"dnsConfig,omitempty"`
+	DNSPolicy *corev1.DNSPolicy    `json:"dnsPolicy,omitempty"`
```
l0kix2, line 291: Since corev1.DNSPolicy is string under the hood and "" value is default zero value in the pod spec I suggest we just use `corev1.DNSPolicy` as type, not pointer, so we don't have to handle 2 empty values.

pkg/components/server.go
```diff
@@ -349,10 +349,16 @@ func (s *serverImpl) rebuildStatefulSet() *appsv1.StatefulSet {
...
 	}
+
 	if ptr.Deref(s.instanceSpec.HostNetwork, s.commonSpec.HostNetwork) {
 		statefulSet.Spec.Template.Spec.HostNetwork = true
-		statefulSet.Spec.Template.Spec.DNSPolicy = corev1.DNSClusterFirstWithHostNet
+		if s.instanceSpec.DNSPolicy != nil {
+			statefulSet.Spec.Template.Spec.DNSPolicy = *s.instanceSpec.DNSPolicy
+		} else {
+			statefulSet.Spec.Template.Spec.DNSPolicy = corev1.DNSClusterFirstWithHostNet
+		}
```
l0kix2, line 361: Current code only allow override DNSPolicy for hostNetwork=true case and I don't see:
1) if there is such a case ever
2) why not to let override for hostNetwork=false.

As far as I understand from https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy if HostNetwork=true — DNSPolicy should be set to DNSClusterFirstWithHostNet.
If this is always true the code should look something like that:
```go 
stsDNSPolicy := s.instanceSpec.DNSPolicy
if ptr.Deref(s.instanceSpec.HostNetwork, s.commonSpec.HostNetwork) {
  // https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy
  stsDNSPolicy = corev1.DNSClusterFirstWithHostNet
} 
statefulSet.Spec.Template.Spec.DNSPolicy = stsDNSPolicy
```

OR 
if there are cases when it is required to set DNSPolicy to the value other than `DNSClusterFirstWithHostNet` we can do set them independently with `DNSClusterFirstWithHostNet` default for hostNetwork case:
```go 
var stsDNSPolicy corev1.DNSPolicy
if ptr.Deref(s.instanceSpec.HostNetwork, s.commonSpec.HostNetwork) {
  // https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy
  stsDNSPolicy = corev1.DNSClusterFirstWithHostNet
}
if s.instanceSpec.DNSPolicy != "" {
  stsDNSPolicy =  s.instanceSpec.DNSPolicy
}
statefulSet.Spec.Template.Spec.DNSPolicy = stsDNSPolicy
```

l0kix2 (changes requested):

pkg/components/server.go
l0kix2: > Isn't my variant more clean, cause it also set HostNetwok=true as it was here initially?

Initially users didn't have the ability to configure DNSPolicy and now they will and it would be discouraging for them to set something like `DNSPolicy=Default` (with default value for `HostNetwork=false`) and this changes wouldn't appear in sts/pods.

> Yes, I have a case when I need to set DNSPolicy=None when HostNEtwork is equal True

let's rewrite like that then:
```go
var stsDNSPolicy corev1.DNSPolicy
if ptr.Deref(s.instanceSpec.HostNetwork, s.commonSpec.HostNetwork) {
  // https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy
  stsDNSPolicy = corev1.DNSClusterFirstWithHostNet
}
if s.instanceSpec.DNSPolicy != "" {
  stsDNSPolicy =  s.instanceSpec.DNSPolicy
}
statefulSet.Spec.Template.Spec.DNSPolicy = stsDNSPolicy
```
That way users can configure hostNetwork and DNSPolicy independently (and we will set reasonable default for hostNetwork=true & DNSPolicy="" as before).

PR #422, opened 2025-01-15
Add auto-buildable update flow

wilwell:
Create auto-buildable flow for components update.

This is pre-requisite for multiple update selectors. Instead of fixed pool of `UpdateFlow` we could construct the flow based on updating components. We could use this functionality after to define more granular updates.

Also in this PR added `UpdatingComponents` as struct containing not only the component type, but also it's name. Fields `Components` and `Flow` are deprecated.

Moreover this PR contains small improvements of setting MonitoringPort. Without that sometimes the operator crashes during the tests execution.


pkg/components/helpers.go
```diff
@@ -138,8 +137,16 @@ func CreateUser(ctx context.Context, ytClient yt.Client, userName, token string,
 }
 
 func IsUpdatingComponent(ytsaurus *apiproxy.Ytsaurus, component Component) bool {
-	componentNames := ytsaurus.GetLocalUpdatingComponents()
-	return (componentNames == nil && component.IsUpdatable()) || slices.Contains(componentNames, component.GetName())
+	components := ytsaurus.GetUpdatingComponents()
+	if components == nil && component.IsUpdatable() {
```
l0kix2, line 141: let's check len(components) just in case, I think it is a best practice not to distinguish nil and empty slices.

controllers/update_flow_steps.go
```diff
@@ -0,0 +1,563 @@
...
+	addImpossibleToStartStep(flow, components)
+	addWaitingForSafeModeEnabledStep(flow, components)
+	addWaitingForTabletCellsSavingStep(flow, components)
+	addWaitingForTabletCellsRemovingStartStep(flow, components)
+	addWaitingForTabletCellsRemovedStep(flow, components)
+	addWaitingForSnapshotsStep(flow, components)
+	addWaitingForPodsRemoval(flow, components)
+	addWaitingForPodsCreation(flow, components)
+	addWaitingForMasterExitReadOnly(flow, components)
+	addWaitingForMasterExitReadOnly(flow, components)
```
l0kix2, line 62: added twice

l0kix2 (commented):
It definitely looks like the move to the direction we've discussed. 
I'm not quite happy with this new flow/check/add-methods implementation, because it is way harder to understand what goes after what what than switch-case before that (of course I understand that this code allows more flexible flow building which isn't possible now and therefore more complex than simple switch). 
Not sure also how to write the test for flow based on different inputs here anyway, since we have functions in stepCheck struct.
Let me think a bit here, maybe I could generate some ideas (considering the issues you've experienced while writing this change) how to make it more easy to read and maintain. Or will have to admit this complexity is inevitable if we want dynamic flow generation.

l0kix2:
Make sure to rebase on main at some point and add new jobs for enable chunks, that I've contributed recently (they shouln't be hard to add to workflow, they are linear and can be chained to master stages).
Also soon new component (kafka proxy) may be [merged](https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/407), so we shouldn't lost it also.

l0kix2:
Again FIY I've added extra steps in flow to support 24.2 update
https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/437
they are still linear though, so adapting shouldn't be problematic

api/v1/ytsaurus_types.go
```diff
@@ -789,17 +791,25 @@ const (
...
 	Flow                  UpdateFlow             `json:"flow,omitempty"`
 	Conditions            []metav1.Condition     `json:"conditions,omitempty"`
 	TabletCellBundles     []TabletCellBundleInfo `json:"tabletCellBundles,omitempty"`
 	MasterMonitoringPaths []string               `json:"masterMonitoringPaths,omitempty"`
 }
 
+type Component struct {
+	ComponentName string               `json:"componentName,omitempty"`
+	ComponentType consts.ComponentType `json:"componentType,omitempty"`
+}
```
l0kix2, line 811: Minor, but I would suggest using Name and Type fields without duplication of the `Component`-part.

controllers/update_flow_steps.go
```diff
@@ -0,0 +1,312 @@
...
+	return &flowTree{
+		index: map[ytv1.UpdateState]*flowStep{
+			head.updateState: head,
+		},
+		head: head,
+		tail: head,
+	}
+}
+
+func (f *flowTree) execute(ctx context.Context, ytsaurus *apiProxy.Ytsaurus, componentManager *ComponentManager) (bool, error) {
```
l0kix2, line 124: I think it would be benefitial to make bool return param named or explain when it is true or false in the comment. Otherwise, you need to check code to understand the sematics.

controllers/update_flow_steps.go
```diff
@@ -0,0 +1,312 @@
+package controllers
```
l0kix2, line 1: Let's maybe split this file into two, one with "flow-framework" code and other with "ytsaurus update flow" code for better readability for our flow building code.

controllers/update_flow_steps.go
```diff
@@ -0,0 +1,312 @@
...
+const (
+	stepResultMarkUnsatisfied stepResultMark = ""
+	stepResultMarkHappy       stepResultMark = "happy"
+	stepResultMarkUnhappy     stepResultMark = "unhappy"
+)
+
+func buildAndExecuteFlow(ctx context.Context, ytsaurus *apiProxy.Ytsaurus, componentManager *ComponentManager, updatingComponents []ytv1.Component) (bool, error) {
+	allComponents := convertToYtComponents(componentManager.allComponents)
+	tree := buildFlowTree(updatingComponents, allComponents)
+	ytsaurus.LogUpdate(ctx, fmt.Sprintf("Update flow starting with %s, updating components: %v, all components: %v", ytsaurus.GetUpdateState(), updatingComponents, allComponents))
```
l0kix2, line 26: This gives a bit sloppy looking log line like
```
2025-02-03T16:10:39.814+0100	DEBUG	events	Update flow starting with WaitingForOpArchiveUpdate, updating components: [], all components: [{ Discovery} { Master} { YtsaurusClient} { DataNode} { DataNode} { HttpProxy} { ExecNode} { TabletNode} { Scheduler} { ControllerAgent}]	{"type": "Normal", "object": {"kind":"Ytsaurus","namespace":"test-e2e-8xkt9","name":"test-ytsaurus","uid":"93971bbb-4102-41eb-9025-7c97ea0ef0e2","apiVersion":"cluster.ytsaurus.tech/v1","resourceVersion":"8127056"}, "reason": "Update"}
```
I suggest we add name in `convertToYtComponents` so they would be distinguishable in case of multiple named components of the same type (I guess we will need it in the future `updateSelectors` change anyway). Also maybe let's add some string representation to the Component type.

pkg/components/helpers.go
```diff
@@ -138,8 +137,16 @@ func CreateUser(ctx context.Context, ytClient yt.Client, userName, token string,
 }
 
 func IsUpdatingComponent(ytsaurus *apiproxy.Ytsaurus, component Component) bool {
-	componentNames := ytsaurus.GetLocalUpdatingComponents()
-	return (componentNames == nil && component.IsUpdatable()) || slices.Contains(componentNames, component.GetName())
+	components := ytsaurus.GetUpdatingComponents()
+	if len(components) == 0 && component.IsUpdatable() {
+		return true
+	}
```
l0kix2, line 143: It may be out of scope of this change, but since we already touching all the places (here and in the component manager basically, where UpdatingComponents is used (and theoretically can break something accidentally and I need to check validness of the diff by looking closely) — maybe we can just get rid of this case where empty/nil list of the components means "update everything"). 

We can just put all the components in that field in case of "full update" and never check for this nil/empty list. That way we can simplify this code, `hasComponent` code (and I guess all flow code, because we don't need to pass allComponents list at all). Also I guess `IsUpdatable()` can be removed (as far as I see it has false only for ytclient, which i don't think even uses `IsUpdatingComponent` function).

So maybe by simplifying this place we can actually reduce possible blast radius of that change since code will be more simple and clear. What do you think?

l0kix2 (commented):
Looks good, I mostly have some nitpicks only.
The main thing I'd like to change is the empty updatingComponents list/passing allComponents to the flow building as I wrote in the comment. Do you think it is possible to remove?

controllers/update_flow_steps.go
wilwell: I've moved couple of methods to the sync.go.

l0kix2 (approved):

pkg/ytconfig/scheduler.go
```diff
@@ -51,7 +51,9 @@ func getControllerAgentLogging(spec *ytv1.ControllerAgentsSpec) Logging {
 func getSchedulerServerCarcass(spec *ytv1.SchedulersSpec) (SchedulerServer, error) {
 	var c SchedulerServer
 	c.RPCPort = consts.SchedulerRPCPort
-	c.MonitoringPort = *spec.InstanceSpec.MonitoringPort
+	if spec.InstanceSpec.MonitoringPort != nil {
+		c.MonitoringPort = *spec.InstanceSpec.MonitoringPort
+	}
```
l0kix2, line 56: It seems there are more places like that which lead to panic and go unnoticed since we running ytop in tests inside kind and it restarts it on panic.
https://github.com/ytsaurus/ytsaurus-k8s-operator/issues/446

In which tests it was reproduced so you want and fix this?

PR #423, opened 2025-01-16
Update flow as a tree

l0kix2:


controllers/flow_tree.go
```diff
@@ -0,0 +1,326 @@
...
+		newConditionalForkStep(
+			ytv1.UpdateStateWaitingForYqlaUpdatingPrepare,
+			// This is the unhappy path.
+			st(ytv1.UpdateStateWaitingForYqlaUpdate),
+			// Happy path will be chained automatically.
+		),
+	).chainIf(
+		updMaster,
+		st(ytv1.UpdateStateWaitingForSafeModeDisabled),
+	)
```
l0kix2, line 322: Here we have full flow build with states in one function, so it is observable/rather easy to change.

controllers/flow_tree.go
```diff
@@ -0,0 +1,326 @@
...
+		updStateless,
+		// This is ugly, I suppose component code can be changed to avoid such transitions.
+		ytv1.UpdateStateWaitingForOpArchiveUpdate,
+		newConditionalForkStep(
+			ytv1.UpdateStateWaitingForOpArchiveUpdatingPrepare,
+			// This is the unhappy path.
+			// it will be chained with the next chained step via deferredChain argument
+			st(ytv1.UpdateStateWaitingForOpArchiveUpdate),
+			// Happy path will be chained automatically.
+		),
```
l0kix2, line 296: In 3 (sch, qt, yqla) places we have this condition check when job may run or may not run, which is very frustrating to implement in the linear style (or i haven't find the good approach yet).

I assume this condition may be incapsulated in the respective components (like OpArchivePrepare and OpArchiveUpdate are just added and if op archive update is not needed — component simply does nothing), so code here can be simplified.

controllers/flow_tree.go
```diff
@@ -0,0 +1,326 @@
...
+
+	// TODO: if validation conditions can be not mentioned here or needed
+	tree.chain(
+		newConditionalForkStep(
+			ytv1.UpdateStatePossibilityCheck,
+			// This is the unhappy path.
+			st(ytv1.UpdateStateImpossibleToStart),
+			// Happy path will be chained automatically.
+		),
+	).chainIf(
```
l0kix2, line 265: The chaining approach allows us to not think about what state will be next when adding it, which simplifies things comparing to the original PR.

controllers/flow_tree.go
```diff
@@ -0,0 +1,326 @@
...
+		}
+		return stepResultMarkUnsatisfied
+	},
+	ytv1.UpdateStateWaitingForPodsCreation: func(ctx context.Context, ytsaurus *apiProxy.Ytsaurus, componentManager *ComponentManager) stepResultMark {
+		if componentManager.allReadyOrUpdating() {
+			return stepResultMarkHappy
+		}
+		return stepResultMarkUnsatisfied
+	},
+}
```
l0kix2, line 245: I've separated flowStep checks in a separate map, so flow building code would be smaller and more observable, but if want it there — we can add couple shortcuts (basically wrapper around bool conditions) and move stuff there (though func signatures are long, not sure if it would be possible to get rid of them).

controllers/flow_tree.go
```diff
@@ -0,0 +1,326 @@
...
+		if ytsaurus.IsUpdateStatusConditionTrue(conditionName) {
+			return stepResultMarkHappy
+		}
+		return stepResultMarkUnsatisfied
+	}
+
+}
+
+type flowStep struct {
+	updateState ytv1.UpdateState
```
l0kix2, line 39: Since all steps have updateState instead of condition func in original PR — we can easily write a test for the resulting flow.

controllers/flow_tree.go
```diff
@@ -0,0 +1,326 @@
...
+	}
+	return current
+}
+
+// for the fast lookup of the step by the update state
+// otherwise we can traverse the graph each time
+type flowTree struct {
+	index map[ytv1.UpdateState]*flowStep
+	head  *flowStep
+	tail  *flowStep
```
l0kix2, line 93: We are storing map state to step (as in original PR) and also have directed graph for steps, which gives us the possibility not to declare next step when adding one.

l0kix2 (commented):
Key takeways:
 1. we can have graph of steps and chain them without specifiying the next step (i suppose if we go with updateSelectors it would be rather painful to build flow as it is done in the original PR)
 2. If update state is used as step id we can write tests for resulting flows.
 3. if flow building code is compact — it is easier to see transitions and change them.

PR #432, opened 2025-01-21
Add location purge key

koct9i:
Changing purge key to different non-empty value
removes all data in location on next start.

### DIFF
--- a/pkg/components/volume.go
+++ b/pkg/components/volume.go
@@ -3,6 +3,7 @@ package components
 import (
 	"fmt"
 	"path"
+	"strings"
 
 	ytv1 "github.com/ytsaurus/ytsaurus-k8s-operator/api/v1"
 	"github.com/ytsaurus/ytsaurus-k8s-operator/pkg/consts"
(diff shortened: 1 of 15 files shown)


pkg/components/volume.go
```diff
@@ -82,11 +83,22 @@ func createServerVolumes(specVolumes []corev1.Volume, configMapName string) []co
...
+		if location.PurgeKey != "" {
+			purgeKeyPath := path.Join(location.Path, consts.PurgeKeyFileName)
+			fmt.Fprintf(&cmd, "[ \"`cat '%s'`\" = '%s' ] || find '%s' -mindepth 1 -delete; ",
+				purgeKeyPath, location.PurgeKey, location.Path)
+			fmt.Fprintf(&cmd, "echo '%s' >'%s'; ", location.PurgeKey, purgeKeyPath)
+		}
 	}
-	return command
+
+	return cmd.String()
```
l0kix2, line 101: Is it possible/hard to write some test that if there are some data in location and purge key is empty — the data is not removed?

koct9i: Golden spec generation should be enough. I'll add it.

Issue #446, opened 2025-02-07
Occasional panic/NPE when components are updating but have no MonitoringPort configured


l0kix2:
It seem to happen in the places like that
https://github.com/ytsaurus/ytsaurus-k8s-operator/blob/984a9192a8e287ce03ca5194087634515ad8dfe3/pkg/ytconfig/query_tracker.go#L24-L24

We are setting mointoringPort in the component constructor, 
https://github.com/ytsaurus/ytsaurus-k8s-operator/blob/984a9192a8e287ce03ca5194087634515ad8dfe3/pkg/components/query_tracker.go#L43-L45

but it seems it got lost, possibly on resource update object is fetched from k8s without monitoring port field.

```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1625742]

goroutine 801 [running]:
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile.func1()
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.18.7/pkg/internal/controller/controller.go:111 +0x1e5
panic({0x1984260?, 0x2ce14d0?})
	/usr/local/go/src/runtime/panic.go:770 +0x132
github.com/ytsaurus/ytsaurus-k8s-operator/pkg/ytconfig.getQueryTrackerServerCarcass(_)
	/workspace/pkg/ytconfig/query_tracker.go:24 +0x82
github.com/ytsaurus/ytsaurus-k8s-operator/pkg/ytconfig.(*Generator).getQueryTrackerConfigImpl(_, _)
	/workspace/pkg/ytconfig/generator.go:736 +0x7f
github.com/ytsaurus/ytsaurus-k8s-operator/pkg/ytconfig.(*Generator).GetQueryTrackerConfig(0xc0011d4a68?, 0xc006ee29a0?)
	/workspace/pkg/ytconfig/generator.go:751 +0x58
github.com/ytsaurus/ytsaurus-k8s-operator/pkg/components.NewQueryTracker.func1()
	/workspace/pkg/components/query_tracker.go:53 +0x25
github.com/ytsaurus/ytsaurus-k8s-operator/pkg/components.(*ConfigHelper).getConfig(0xc006ee2b00, {0x1bf8e84, 0x1b})
[...]

PR #447, opened 2025-02-07
Fix zero port if no monitoring port configured

l0kix2:
@wilwell I've reverted your fix about setting nils in monitoring from https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/422
This places are not panicing now, but they make things worse in some sense: monitoring port for components sometimes set 0, which leads for example to successfully updated, but non ready master.

This issue should be understood better and fixed properly for all components here https://github.com/ytsaurus/ytsaurus-k8s-operator/issues/446

UPD. Ended with explicitly setting default value in all component configs, because it is faster and simpler than reproducing the bug which panics.
Removed overrides from all places in cfg tests, no canondata changed as expected, added separate test for master with port override.


pkg/ytconfig/yql_agent.go
```diff
@@ -50,7 +50,10 @@ func getYQLAgentLogging(spec *ytv1.YQLAgentSpec) Logging {
 func getYQLAgentServerCarcass(spec *ytv1.YQLAgentSpec) (YQLAgentServer, error) {
 	var c YQLAgentServer
 	c.RPCPort = consts.YQLAgentRPCPort
-	c.MonitoringPort = *spec.InstanceSpec.MonitoringPort
+	c.MonitoringPort = consts.YQLAgentMonitoringPort
```
koct9i, line 53: This could be one line:
```c.MonitoringPort = ptr.Deref(spec.InstanceSpec.MonitoringPort, consts.YQLAgentMonitoringPort)```

**Comment by koct9i** [comment]
also please remove piieces like 
```
	if spec.InstanceSpec.MonitoringPort == nil {
		spec.InstanceSpec.MonitoringPort = ptr.To(int32(consts.ExecNodeMonitoringPort))
	}
```
pkg/components/exec_node.go and other places - reconsiller shouldn't change spec

l0kix2:
> also please remove piieces like
> 
> ```
> 	if spec.InstanceSpec.MonitoringPort == nil {
> 		spec.InstanceSpec.MonitoringPort = ptr.To(int32(consts.ExecNodeMonitoringPort))
> 	}
> ```
> 
> pkg/components/exec_node.go and other places - reconsiller shouldn't change spec

let's merge that one since tests are green and change is independent and I fix this thing in https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/448 after rebase

PR #449, opened 2025-02-12
Add lost CA bundle and TLS secrets VolumeMounts for jobs container

imakunin:


pkg/components/exec_node_base.go
```diff
@@ -177,6 +179,8 @@ func (n *baseExecNode) doBuildCRISidecar(envSpec *ytv1.JobEnvironmentSpec, podSp
 	}
 
 	podSpec.Containers = append(podSpec.Containers, jobsContainer)
+
+	return &podSpec.Containers[len(podSpec.Containers)-1]
```
l0kix2, line 183: Why not simply `return &jobsContainer`?

l0kix2 (commented):
I'm not happy overall we have this server _interface_ inside components and we need to pass stuff inside and out, but that's a different story

pkg/components/exec_node_base.go
imakunin: I'm not a real go-developer :).
My C++ background warns me against passing pointers to internal objects outside.
I thought about it, but I didn't test it, so that's it.

l0kix2: Yes, you can do it, it is safe in golang, it manages allocating memory (on heap I suppose, but I'm not big specialist) for you in that case and gc will collect it when no references left.
It actually a best practice in golang, almost all the code that creates some struct in constructor-like fashion returns pointer in the return value.

psushin: You could add all the necessary mounts inside this function, since you already have access to ```n.server```. Than you wouldn't have to change the function signature.

l0kix2 (approved):

PR #466, opened 2025-03-13
Set init quota for user-defined mediums

futujaos:
After I set non-default mediums for tablet cell bundles, some components (scheduler, query tracker, queue agent, strawberry) run operations in corresponding tablet nodes during cluster initialization, but, unfortunately, there is no quota set for new mediums (only quota for `default` medium with value ~1Tb). 

So here I suggest to set disk space quota for sys account for new mediums, enough for cluster initialization purposes.

### DIFF
--- a/pkg/components/master.go
+++ b/pkg/components/master.go
@@ -19,6 +19,7 @@ import (
 
 const (
 	defaultHostAddressLabel = "kubernetes.io/hostname"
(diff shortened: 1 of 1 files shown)


psushin (approved):
LGTM - this is crucial to start tablet cells with overriden mediums.
And actual account limits anyway should be set separately, after the cluster has started.

PR #468, opened 2025-03-17
tablet node tag filter for bundles in bootstrap

futujaos:
Currently there is no convenient way to attach bundles to specific tablet node group.

Combining:
- `ClusterNodesSpec.Tags` in `ytv1.TabletNodesSpec` 
- `BundleBootstrapSpec.NodeTagFilter`

we can attach bundle to tablet node group using tags.


pkg/components/tablet_node.go
```diff
@@ -205,6 +205,20 @@ func (tn *TabletNode) initBundles(ctx context.Context) (ComponentStatus, error)
...
 	}
 
+	for _, bundle := range []string{DefaultBundle, SysBundle} {
+		bootstrap := tn.getBundleBootstrap(bundle)
+		nodeTagFilter := ""
+		if bootstrap != nil && bootstrap.TabletNodeTagFilter != nil {
+			nodeTagFilter = *bootstrap.TabletNodeTagFilter
+		}
+		path := ypath.Path("//sys/tablet_cell_bundles").Child(bundle).Attr("node_tag_filter")
+		err = ytClient.SetNode(ctx, path, nodeTagFilter, nil)
```
futujaos, line 215: always settings `node_tag_filter` seems ok even if filter is `""`
- if bundles initialized for the first time, attr will not be set
- if bundles was already initialized, we will drop existing filter as intended by user

achulkov2: Ok, I see your point.

We definitely cannot have modifications of `node_tag_filter` when they were not explicitly requested, it will break user clusters. 
Forcing administrators to unset it manually is the better (or maybe only) way.

pkg/components/tablet_node.go
```diff
@@ -205,6 +205,23 @@ func (tn *TabletNode) initBundles(ctx context.Context) (ComponentStatus, error)
 		}
 	}
 
+	for _, bundle := range []string{DefaultBundle, SysBundle} {
+		bootstrap := tn.getBundleBootstrap(bundle)
+		nodeTagFilter := ""
```
l0kix2, line 210: this is unused one and ends up in a log, which is misleading, because different value is being set.

l0kix2 (changes requested):
Other than that LGTM, but I want final ok from @achulkov2 who understands the logic behind node_tag_filter

api/v1/ytsaurus_types.go
```diff
@@ -194,7 +194,8 @@ type BundleBootstrapSpec struct {
 	SnapshotPrimaryMedium  *string `json:"snapshotMedium,omitempty"`
 	ChangelogPrimaryMedium *string `json:"changelogMedium,omitempty"`
 	//+kubebuilder:default:=1
-	TabletCellCount int `json:"tabletCellCount,omitempty"`
+	TabletCellCount     int     `json:"tabletCellCount,omitempty"`
+	TabletNodeTagFilter *string `json:"tabletNodeTagFilter,omitempty"`
```
achulkov2, line 198: Let's drop `Tablet` so that the name of the variable corresponds to the name of the attribute being set.
(there is no point in filtering any other nodes beside tablet nodes for tablet cell bundles, so it is irrelevant)

achulkov2 (approved):
One renaming request, otherwise LGTM.

l0kix2 (approved):

PR #475, opened 2025-04-01
Configure yqla mrjob syslibs V2

Krisha11:
Improvement of https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/409/files

### DIFF
--- a/api/v1/ytsaurus_types.go
+++ b/api/v1/ytsaurus_types.go
@@ -563,11 +563,6 @@ type StrawberryControllerSpec struct {
 
 type YQLAgentSpec struct {
 	InstanceSpec `json:",inline"`
-
-	// Sets mrjob dynamic library paths in yql agent config.
-	//+kubebuilder:default:=false
-	//+optional
-	ConfigureMrJobSystemLibs bool `json:"configureMrJobSystemLibs,omitempty"`
 }
 
 type QueueAgentSpec struct {
--- a/config/crd/bases/cluster.ytsaurus.tech_ytsaurus.yaml
+++ b/config/crd/bases/cluster.ytsaurus.tech_ytsaurus.yaml
@@ -40317,10 +40317,6 @@ spec:
                             x-kubernetes-list-type: atomic
                         type: object
                     type: object
-                  configureMrJobSystemLibs:
-                    default: false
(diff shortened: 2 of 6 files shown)


l0kix2:
@Krock21 do you want to review this PR?

Krock21:
Did we have a release with `configureMrJobSystemLibs`? If yes, then I think we should think about existing ytsaurus specs that specify `configureMrJobSystemLibs`
Am I right that they will fail with validation error?

LGTM otherwise

Krock21 (approved):

Krock21:
Users may be surprised if their specs with `configureMrJobSystemLibs` will start failing

Krisha11:
> Did we have a release with configureMrJobSystemLibs? If yes, then I think we should think about existing ytsaurus specs that specify configureMrJobSystemLibs
Am I right that they will fail with validation error?

We haven't any qt release requires this option. 

> Users may be surprised if their specs with `configureMrJobSystemLibs` will start failing

Only nebius knows about this options, I suppose you can tell others about this changes.
Also I am going to put it in k8s release notes

l0kix2:
> I don't know if we have users who track releases and tried it
As far as I understood, nothing should change for users who set it to true, because not it is true unconditionally. And for those who explicitly set false (which is default) we now simply removing such possibility and I suppose it should be needed for old and new images.

> Did we have a release with configureMrJobSystemLibs? If yes, then I think we should think about existing ytsaurus specs that specify configureMrJobSystemLibs
> Am I right that they will fail with validation error?

Since field is non-nullable, there is a field in existing installed CR's (either false or true). And also there is a field in yaml specs users maintain.

I assume no one really used the field in spec (and if validation fails on the next apply I think they can easily figure it out based on release notes), so we only should care about installed CR's with default or non-default value — do the validation will happen on helm chart/crd update?

According to chat gpt:
>	•	If preserveUnknownFields: false is set (default in Kubernetes 1.22+), the old field is dropped when the CR is saved.
So as far as I understand, for new k8s it should be simply removed.

I've tried to upgrade CRD (to the one with removed field) in a local cluster and haven't got any error even providing a field explicitly in my spec.

I think we can roll it with a release note, but it would be nice to have an understanding of that behaviour and put in our guidlines.

PR #479, opened 2025-04-11
Rerun init queue agent script during QA update

savnadya:


pkg/components/queue_agent.go
```diff
@@ -316,12 +329,88 @@ func (qa *QueueAgent) prepareInitQueueAgentState() {
...
+}
+
+func (qa *QueueAgent) initQAState(ctx context.Context, dry bool) (ComponentStatus, error) {
+	for _, tnd := range qa.tabletNodes {
+		tndStatus, err := tnd.Status(ctx)
+		if err != nil {
+			return tndStatus, err
+		}
+		if !IsRunningStatus(tndStatus.SyncStatus) {
+			// Wait for tablet nodes to proceed with operations archive init.
```
l0kix2, line 367:
```suggestion
			// Wait for tablet nodes to proceed with queue agent init.
```
This was a bit misleading copy-paste, I started to think how we can wait for op archive by checking tnds.

l0kix2 (changes requested):
LGTM overall, could you though please check if any of basic e2e, which update cluster, have QA in the test spec and if not — please add it it, so we have some coverage for that piece of flow and qa init job.

l0kix2:
Tests will need to be fixed a little, since they are checking what components/pods/jobs were involved.

l0kix2 (approved):

PR #480, opened 2025-04-21
sync exec node if containerd config changes

sanchosancho:


psushin:
LGTM

l0kix2:
LGTM overall, but I think test won't hurt indeed. e2e are long and it is bit frustrating to make them longer, but we already have similar test case about config overrides triggering update, so we can just extend it to test that sidecar spec change would trigger end's update https://github.com/ytsaurus/ytsaurus-k8s-operator/blob/c3eba18aa6d482e8a56c54d5b697c4b83840fe47/test/e2e/ytsaurus_controller_test.go#L523-L523

sanchosancho:
Sorry, being a bit busy lately :)
Finally modified the suggested test by adding containerd.toml override. It looks fine in local env.

l0kix2 (approved):

PR #520, opened 2025-07-29
Pass metadata env vars into server pod containers

Gufran:
Add a new environment variable `K8S_POD_NAME` in exec node pod. The value of the variable is set from `metadata.name` using downwardAPI


koct9i (approved):
Ok. This will change all pods. But our tests does not detect that. And I see no other option and reason why make this configurable right now. Eventually we could make environment variables configuration explicit, including downward. Right now I don't care.

l0kix2 (approved):

PR #531, opened 2025-08-15
Introduce cypress patches

koct9i:
- **go.mod: upgrade to go 1.23.12**
- **Add cypress yson patch**
- **Add update phase for applying per-component cypress patches**
- **Add patch for primary masters in cluster connection**


controllers/component_manager.go
```diff
@@ -39,13 +39,19 @@ func NewComponentManager(
...
+	}
+
 	d := components.NewDiscovery(cfgen, ytsaurus)
 	m := components.NewMaster(cfgen, ytsaurus)
 	var hps []components.Component
 	for _, hpSpec := range ytsaurus.GetResource().Spec.HTTPProxies {
 		hps = append(hps, components.NewHTTPProxy(cfgen, ytsaurus, m, hpSpec))
 	}
-	yc := components.NewYtsaurusClient(cfgen, ytsaurus, hps[0])
+	yc := components.NewYtsaurusClient(cfgen, ytsaurus, hps[0], getAllComponents)
```
l0kix2, line 54: Have you considered to have different virtual component like cypressPatcher or something, so we don't have those weird cyclic dependencies between components and yt client?

l0kix2 (approved):

PR #543, opened 2025-08-27
Get or guess cluster domain only once at start

koct9i:
Guessing every time and failing back to default when DNS fails is a bad idea.

Signed-off-by: Konstantin Khlebnikov <khlebnikov@tracto.ai>


koct9i:
it still here

kruftik (approved):

PR #547, opened 2025-08-27
Nvidia GPU support: nvidia-container-runtime & entrypoint for jobs container with GpuAgent

futujaos:
This PR is implementation of https://github.com/ytsaurus/ytsaurus-k8s-operator/discussions/533

For out-of-box support of NVidia GPU we need to do the following for exec nodes with GPU:
- Set `nvidia-container-runtime` as containerd runtime
- Run [GpuAgent](https://github.com/ytsaurus/ytsaurus/tree/main/yt/yt/gpuagent) together with containerd (o cri-o) in jobs container. Exec node will request information about GPU from this agent. To achieve this, we set `jobs_entrypoint.sh` with its arguments as command for jobs container. `jobs_entrypoints.sh` will be the part of main YTsaurus docker image, as well as `nvidia-container-runtime`
- `ExecNodesSpec.JobEnvironment.Runtime` section is added to YTsaurus spec with option to set NVidia Runtime.


pkg/ytconfig/cri.go
```diff
@@ -18,6 +18,7 @@ type CRIConfigGenerator struct {
 	Isolated       bool
 	StoragePath    *string
 	MonitoringPort int32
+	HasGPU         bool
```
imakunin, line 21: I would suggest calling it smth like `autoConfigureGPU` to explicitly indicate that there will be an attempt to do something automatically (possibly unsuccessful, if we have not nvidia device)

pkg/ytconfig/cri.go
```diff
@@ -144,3 +140,45 @@ func (cri *CRIConfigGenerator) GetContainerdConfig() ([]byte, error) {
 	// TODO(khlebnikov): Refactor and remove this mess with formats.
 	return marshallYsonConfig(config)
 }
+
+func (cri *CRIConfigGenerator) defineContainerdRuntimes() (runtimes map[string]any, defaultRuntimeName string) {
```
imakunin, line 144: s/define/create

pkg/ytconfig/cri.go
```diff
@@ -144,3 +140,45 @@ func (cri *CRIConfigGenerator) GetContainerdConfig() ([]byte, error) {
...
+}
+
+func execNodesRequestsGPU(spec *ytv1.ExecNodesSpec) bool {
+	if spec.JobResources == nil {
+		return false
+	}
+	return resourceListHasGPU(spec.JobResources.Requests) || resourceListHasGPU(spec.JobResources.Limits)
+}
+
+func resourceListHasGPU(resourceList corev1.ResourceList) bool {
```
imakunin, line 177: maybe `requiresNvidiaGPU` ?

koct9i: Please add explicit spec section to runtime configuration next to "cri" with optional field "nvidia" which could empty struct but I guess you already have parameters in mind.

```
jobEnvironment:
  cri:
    ...
  runtime:
    nvidia:
     ...

```

futujaos: explicit section required API change, however, exec nodes with nvidia GPU have no other way to work except this `containerd.toml` patch, and doing it automatically shouldn't break anything on existing clusters

do we really need an explicit config section?

futujaos: @koct9i 

```yaml
jobEnvironment:
  cri:
    ...
  runtime:
    nvidia:
     ...
```

Explicit section – ok, but we still want to provide user with simple setup, otherwise users will forget to specify runtime for GPU nodes.

What if we auto-detect (by job resources, like now) runtime in case if user didn't specify `runtime` section? And user will still have option to manually specify runtime in case if we didn't take some cases into account.

```go
type JobEnvironmentSpec struct {
  ...
  // CRI runtime, auto-detect by default (nil)
  //+optional
  Runtime *JobRuntimeSpec
}

type JobRuntimeSpec struct {
  // CRI job runtime: runc (default), nvidia
  //+optional
  //+kubebuilder:validation:Enum={"runc","nvidia"}
  Type *JobRuntimeType
}

type JobRuntimeType string

const (
  JobRuntimeRunc   JobRuntimeType = "runc"
  JobRuntimeNvidia JobRuntimeType = "nvidia"
  ...
)
```

(in this example I specified `runc` as default runtime type, but I don't know if it is the case for CRI-O, may be it should be `default` option)

koct9i: Is this user who cannot comprehend such setup in the room with you right now? This system is already complicated enough. There are tons of parameters which need to be adjusted consistently.

Default behavior must be as simple as possible.

All non-trivial features must be explicit.

If you want to make any inconsistency clear -- you could add validation and generate error or warning. I prefer warning because this is non-fatal misconfiguration and I foresee use cases for that.

api/v1/ytsaurus_types.go
```diff
@@ -493,6 +501,9 @@ type JobEnvironmentSpec struct {
 	// Do not use slot user id for running jobs.
 	//+optional
 	DoNotSetUserId *bool `json:"doNotSetUserId,omitempty"`
+	// Define custom job runtime.
+	//+optional
+	Runtime *JobRuntimeSpec `json:"runtime,omitempty"`
```
imakunin, line 506: nvidia-runtime is relevant only for the container scheme

Maybe it should be moved to the CRI settings?

futujaos: i made it as @koct9i suggested:
```
jobEnvironment:
  cri:
    ...
  runtime:
    nvidia:
     ...
```

imakunin: Ok, but it seems to me that there is a messy naming here. 
What about `JobEnvironmentRuntime` instead of `JobRuntimeSpec`?

futujaos: `Runtime` is located inside `JobEnvironmentSpec`, so it's clear from context. `CRI` named in the same way, so it's consistent with existing naming.

PR #553, opened 2025-09-05
Support CHYT's server in http-proxies

epsilond1:
Support the ability to run a proxy server for CHYT on http proxies


api/v1/ytsaurus_types.go
```diff
@@ -355,6 +355,9 @@ type HTTPProxiesSpec struct {
 	HttpsPort     *int32 `json:"httpsPort,omitempty"`
 	HttpNodePort  *int32 `json:"httpNodePort,omitempty"`
 	HttpsNodePort *int32 `json:"httpsNodePort,omitempty"`
+	//+kubebuilder:default:=8123
+	ChytHttpServerPort     *int32 `json:"chytHttpServerPort,omitempty"`
+	ChytHttpServerNodePort *int32 `json:"chytHttpServerNodePort,omitempty"`
```
koct9i, line 360: Can we add https too?

Maybe it would be better to group options for http server into structure and use it embedded and as "chyt" field:

```
 httpPort: ..
 httpsPort: ..
 chyt:
   httpPort: ..
   httpsPort ..
```

koct9i: Yeah. https is already here

```
    NHttp::TServerConfigPtr HttpServer;
    NHttps::TServerConfigPtr HttpsServer;
    NHttp::TServerConfigPtr TvmOnlyHttpServer;
    NHttps::TServerConfigPtr TvmOnlyHttpsServer;
    NHttp::TServerConfigPtr ChytHttpServer;
    NHttps::TServerConfigPtr ChytHttpsServer;
```

pkg/ytconfig/proxy.go
```diff
@@ -130,6 +131,9 @@ func getHTTPProxyServerCarcass(spec *ytv1.HTTPProxiesSpec) (HTTPProxyServer, err
 	c.RPCPort = consts.HTTPProxyRPCPort
 	c.MonitoringPort = ptr.Deref(spec.InstanceSpec.MonitoringPort, consts.HTTPProxyMonitoringPort)
 	c.Port = int(ptr.Deref(spec.HttpPort, int32(consts.HTTPProxyHTTPPort)))
+	c.ChytHttpServer = &HTTPServer{
```
koct9i, line 134: Let's add them only if http proxy spec has "chyt" field.

l0kix2:
Is it an optional feature? If so, let's find a way to distinguish enabled/disabled modes in the ytsaurus spec and only add all extra configurations in case if feature is enabled.

api/v1/ytsaurus_types.go
epsilond1: I fixed it, but chyt -> chytServer, is it ok?

pkg/ytconfig/proxy.go
epsilond1: Is this such solution that you means? Anyway, we have default ports for http(s).

koct9i (changes requested):
Need to handle nil correctly
and rename "chytSever" into "chytProxy" or something.

pkg/ytconfig/proxy.go
```diff
@@ -113,6 +115,29 @@ func getHTTPProxyLogging(spec *ytv1.HTTPProxiesSpec) Logging {
...
+					FileName: path.Join(consts.HTTPSSecretMountPoint, corev1.TLSCertKey),
+				},
+				PrivateKey: PemBlob{
+					FileName: path.Join(consts.HTTPSSecretMountPoint, corev1.TLSPrivateKeyKey),
+				},
+				UpdatePeriod: yson.Duration(consts.HTTPSSecretUpdatePeriod),
+			},
+		}
+	}
+}
```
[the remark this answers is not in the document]
koct9i: No. Let's enable only when explicitly enabled in spec.

I though about per-proxy group setup:
```
spec:
 httpProxies
  - name: default
    chytProxy: {}
```

But there is another option for "cluster features".
for example something like:
```
spec:
 clusterFeatures:
  httpProxyHaveClickHouseAddress: true
```

pkg/components/httpproxy.go
```diff
@@ -32,6 +32,44 @@ func NewHTTPProxy(
...
+		chytProxy, err = ytconfig.MakeChytProxySpec(spec)
+		if err != nil {
+			return nil
+		}
+	}
+
+	if chytProxy != nil {
+		containerPorts = append(containerPorts, corev1.ContainerPort{
+			Name:          consts.CHYTHttpProxyName,
+			ContainerPort: ptr.Deref(chytProxy.HttpPort, int32(consts.HTTPProxyChytHttpPort)),
```
epsilond1, line 64: Maybe `ptr.Deref(...)`is not needed here? @koct9i

koct9i: Ah. That not chyt-proxy spec form spec but "spec" from config.

koct9i: I see no reason for ytconfig.MakeChytProxySpec
that could be written in place sorter and clearer

```
chytProxy := ptr.Deref(httpProxy.ChytProxy, ytv1.CHYTProxySpec{})
httpPort := ptr.Deref(chytProxy.HttpPort, consts.ChytDefaultHttpPort)
httspPort := ptr.Deref(chytProxy.HttpsPort, consts.ChytDefaultHttpsPort)
```

checking for zero port at this stage is redundant, such checks should be in validator

koct9i (changes requested):

pkg/testutil/spec_builders.go
```diff
@@ -120,6 +120,9 @@ func (b *YtsaurusBuilder) CreateMinimal() {
 				UseShortNames:    true,
 				CoreImage:        b.YtsaurusImage,
 				JobImage:         b.JobImage,
+				ClusterFeatures: &ytv1.ClusterFeatures{
+					HTTPProxyHaveChytAddress: true,
```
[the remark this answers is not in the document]
koct9i: We have "minimal" as basis for all test setup.
feature-specific setup for feature tests
and "all-components" for maximum coverage

pkg/components/httpproxy.go
epsilond1: And about zero, see comment: https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/553#discussion_r2340654056, pt. 2

koct9i (approved):

PR #567, opened 2025-10-01
Add OffshoreDataGateway component

pavel-bash:
# Description

This PR introduces an OffshoreDataGateway component. It does not yet exist in the open-source code of YT, but will eventually be merged.

### DIFF
--- a/api/v1/offshoredatagateways_types.go
+++ b/api/v1/offshoredatagateways_types.go
@@ -39,6 +39,9 @@ type OffshoreDataGatewaysStatus struct {
 //+kubebuilder:subresource:status
 
 // OffshoreDataGateways is the Schema for the OffshoreDataGateways API
+// Be careful: this component is experimental and is not part of the public
+// API yet, so there are no guarantees it will work even when configured
+// TODO(pavel-bash): remove this warning when the component is released to open-source
 type OffshoreDataGateways struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
--- a/pkg/consts/types.go
+++ b/pkg/consts/types.go
@@ -18,20 +18,17 @@ const (
 	SchedulerType            ComponentType = "Scheduler"
 	StrawberryControllerType ComponentType = "StrawberryController"
 	TabletNodeType           ComponentType = "TabletNode"
-	// OffshoreDataGateway is experimental and is not a part of the public
-	// API yet, so there are no guarantess it will work when configured
-	// TODO(pavel-bash): remove this warning when the component is released to open-source
-	OffshoreDataGatewayType ComponentType = "OffshoreDataGateway"
-	TcpProxyType            ComponentType = "TcpProxy"
-	KafkaProxyType          ComponentType = "KafkaProxy"
-	UIType                  ComponentType = "UI"
-	YqlAgentType            ComponentType = "YqlAgent"
-	YtsaurusClientType      ComponentType = "YtsaurusClient"
-	ChytType                ComponentType = "CHYT"
-	SpytType                ComponentType = "SPYT"
-	ClusterConnectionType   ComponentType = "ClusterConnection"
-	NativeClientConfigType  ComponentType = "NativeClientConfig"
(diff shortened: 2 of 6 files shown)


pkg/components/offshore_data_gateway.go
```diff
@@ -0,0 +1,83 @@
...
+type OffshoreDataGateway struct {
+	server server
+	cfgen  *ytconfig.NodeGenerator
+	spec   *ytv1.OffshoreDataGatewaySpec
+	baseComponent
+}
+
+func NewOffshoreDataGateways(
+	cfgen *ytconfig.NodeGenerator,
+	nodes *ytv1.OffshoreDataGateways,
```
l0kix2: gateways?
unused?

pkg/ytconfig/offshore_data_gateways.go
```diff
@@ -0,0 +1,30 @@
...
+
+func getOffshoreDataGatewaysLogging(spec *ytv1.OffshoreDataGatewaySpec) Logging {
+	return createLogging(
+		&spec.InstanceSpec,
+		"offshore-data-gateway",
+		[]ytv1.TextLoggerSpec{defaultInfoLoggerSpec(), defaultStderrLoggerSpec()})
+}
+
+func getOffshoreDataGatewaysCarcass(spec *ytv1.OffshoreDataGatewaySpec) (OffshoreDataGatewayServer, error) {
+	var onps OffshoreDataGatewayServer
```
l0kix2:
```suggestion
	var odgs OffshoreDataGatewayServer
```
:)

api/v1/offshoredatagateways_types.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// OffshoreDataGatewaysSpec defines the desired state of OffshoreDataGateways
+type OffshoreDataGatewaysSpec struct {
+	RemoteClusterSpec       *corev1.LocalObjectReference `json:"remoteClusterSpec"`
+	CommonSpec              `json:",inline"`
+	OffshoreDataGatewaySpec `json:",inline"`
```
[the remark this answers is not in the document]
pavel-bash: > I would suggest just start simple and embed InstanceSpec here directly to have less confusions

I really wanted to do that as well. However, some fields are repeated in `InstanceSpec` and in `CommonSpec`; for instance, `hostNetwork` and `nativeTransport`. This prevents the code from compiling as those fields become ambiguous due to both being `json:",inline"`.

I don't know why this error is not thrown now though, considering that `InstanceSpec` is also inlined in `OffshoreDataGatewaySpec`, and `OffshoreDataGatewaySpec` is inlined in `OffshoreDataGatewaysSpec`... maybe the compiler cannot look at 2+ layers of inline.

Anyway, I also would love to get rid of the second spec, but for that we'll need to make the `InstanceSpec` here non-inline, and I haven't found any examples of that in the code, it's always inlined. But if you say so, I'll do it.

pkg/consts/types.go
```diff
@@ -18,16 +18,20 @@ const (
...
-	YqlAgentType             ComponentType = "YqlAgent"
-	YtsaurusClientType       ComponentType = "YtsaurusClient"
-	ChytType                 ComponentType = "CHYT"
-	SpytType                 ComponentType = "SPYT"
-	ClusterConnectionType    ComponentType = "ClusterConnection"
-	NativeClientConfigType   ComponentType = "NativeClientConfig"
-	TimbertruckType          ComponentType = "Timbertruck"
+	// OffshoreDataGateway is experimental and is not a part of the public
+	// API yet, so there are no guarantess it will work when configured
+	// TODO(pavel-bash): remove this warning when the component is released to open-source
```
l0kix2: I probably haven't made myself clear enough, I've meant let's add the comment in 
api/v1/offshoredatagateways_types.go
because no one will find it here, but if someone interested in component — he would check the main struct maybe. Also the comment should also be replicated in markdown docs from there (I think)

l0kix2 (approved):
Let's wait for the tests and merge

Issue #374, opened 2024-10-25
Detect all diffs for components including sts fields


koct9i:
### Discussed in https://github.com/ytsaurus/ytsaurus-k8s-operator/discussions/366

Instead of hardcoded "need sync" checks operator must simply look at resource generation and
for all changes regenerate all dependent resources to compare them with present state.

l0kix2:
I suggest we discuss our options, compare and choose the one which most fit us.

We have one implementation with field-by-field compare https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/573
But I think we should research generation/hash annotations patter or maybe server side updates (though it is probably required compatible version of k8s, which I'm not sure we can depend on).

koct9i:
Note: also for timbertruck sidecar

https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/812

PR #573, opened 2025-10-14
Made operator update StatefulSets not only when replica count changed

qurname2:
PR is aiming to solve Issue #374.
**Overview**
This PR implements comprehensive StatefulSet spec change detection to enable the operator to detect and respond to configuration changes (resources, env vars, volumes, etc.) beyond just replica count changes and images.

**Changes:**

1. Field-by-Field Pod Template Comparison, [link](https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/573/files#diff-b4dbab1d61d27d5d4d375712a2fa54d744179dd074209d9f117079ee8a3f697f)
That logic compares:
- Containers & InitContainers: Image, command, args, env vars, resources, volume mounts
- Volumes: ConfigMap, Secret, PersistentVolumeClaim, EmptyDir, HostPath
- Scheduling: Node selectors, tolerations, affinity rules
- Service Account: Service account name
I tried to implement comparison using **k8s.io/apimachinery/pkg/api/equality** package, but it led to the infinite recreating loop. The same applies to the [strategicpatch.CreateTwoWayMergePatch](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/strategicpatch#CreateTwoWayMergePatch) (it uses json document merging, not semantic equality, but the problem is similar)
Both of them detects Kubernetes default values, that doesn't matter for us to compare, but led to the false-positive Update status on the operator. 
Here you can see how postgres-operator compares sts's fields: https://github.com/zalando/postgres-operator/blob/3bc244fe397ad084de8ddc3e02a1670f6b4c0949/pkg/cluster/cluster.go#L648
tl;dr - also field-by-field

2. Component-Specific Modifications

- [Fixed](https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/573/files#diff-e8c177ec5552f16b5c1d0b28e41d7e3f6023672e05ad86de72535bd8dcad4ced) YqlAgent false positive detection.

Problem: YqlAgent adds environment variables (YT_FORCE_IPV4, YT_FORCE_IPV6, token) and modifies commands AFTER building the base spec, causing mismatches during comparison
Solution: Extracted modifications into [applyYqlAgentContainerModifications](https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/573/files#diff-e8c177ec5552f16b5c1d0b28e41d7e3f6023672e05ad86de72535bd8dcad4ced)()
Overrode needUpdate() and needStatefulSetSpecUpdate() to apply modifications before comparison

- [execNode](https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/573/files#diff-0f5dd8e5febc9d956f0cc8b083df1092df5e1371cab2d60fd3902adeb4239ecaR65) - ExecNode combines NodeResources + JobResources, requiring special handling. Overrode needStatefulSetSpecUpdate() to compute expected total resources before comparison

- [httProxy](https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/573/files#diff-aa650d84e19619011d241fcf639d366c8362ba5c054713b92a96390319387092R130) - because httpProxy adds volumes for TLS encryption after the buildStatefulSet() and it will cause infinite update loop, cause during every reconcilation loop operator will think there is difference in the statefulSet spec.

- [rpcProxy](https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/573/files#diff-c567a8971a9b57af1d5f56030c211f61fc7e89d361c04db925bbda3146764bd3R120) - the same as for http-proxy

3. Server Infrastructure Updates: [pkg/components/server.go](https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/573/files#diff-1bc37670cc1770517eef105d766108d84bb28c2e46e31f8b4937eb8a44cef316)
Added [needUpdateWithSpecCheck](https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/573/files#diff-1bc37670cc1770517eef105d766108d84bb28c2e46e31f8b4937eb8a44cef316R283)() to allow components to provide custom spec comparison logic.
Updated [needUpdate](https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/573/files#diff-1bc37670cc1770517eef105d766108d84bb28c2e46e31f8b4937eb8a44cef316R303) to use the new comparison function
4. E2E Tests ([test/e2e/ytsaurus_controller_test.go](https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/573/files#diff-382513c8c2d6af42792c49c398acebb81bb621bdfb797022ec14abfbe7af14db))
Added several tests:
- Discovery Resource Update: Tests CPU/memory limit changes with proper UpdateSelector set.
- Discovery Environment Variables Update: Verifies env var changes trigger updates with proper UpdateSelector set.
- Discovery nodeSelector update: Verifies nodeSelector will be updated with proper UpdateSelector set.
- Blocked Updates: Validates that UpdatePlan correctly blocks updates when configured


koct9i:
Maybe we could completely eliminate content comparison of managed resources.

"Controller" resorce (ytsaurus) has generation which is updated at any changes.
We could save this generation into managed resources metadata.

We just compare  controller resource generation and generation saved in managed resource.
At difference we generate new managed resource and apply.
If nothing have changed -> next level controller will do nothing.

PR #576, opened 2025-10-21
Adding customization of the solomon_exporter key

kmalov:
In some cases, you need to change or expand the list of instance tags for solomon_exporter.
In the component specification, instead of using the name solomon_exporter, I use metrics_exporter, as it is more understandable for external users. But if it adds confusion, then we can roll back to solomon_exporter.


achulkov2 (changes requested):

pkg/ytconfig/generator.go
```diff
@@ -246,11 +247,37 @@ func (g *NodeGenerator) fillAddressResolver(c *AddressResolver) {
 	c.Retries = &retries
 }
 
-func (g *NodeGenerator) fillSolomonExporter(c *SolomonExporter) {
+func (g *NodeGenerator) fillSolomonExporter(c *SolomonExporter, m *ytv1.MetricsExporter) {
```
verytable: "c" in `c *SolomonExprorter` looks a bit strange

s might be better

PR #603, opened 2025-12-03
Validate update plan consistency

koct9i:
- component types and class should be valid
- plan with class Nothing or Everything should not contain anything else

Signed-off-by: Konstantin Khlebnikov <khlebnikov@tracto.ai>


qurname2 (approved):
rebased from main, LGTM.

PR #606, opened 2025-12-04
Added HasCustomUpdateState in order to migrate qt,qa,yqla and sch to the common handleUpdatingClusterState

qurname2:
In that [PR](https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/601) I created a bug. qt, qa, yqla and sch components have their own additional logic during the update. [updateQTState](https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/606/files#diff-8de26e1e52d4bdfeb4d2cea625c03860851a33787654f5951ea7719df9896adbR104) for qt, [updateYqla](https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/606/files#diff-e8c177ec5552f16b5c1d0b28e41d7e3f6023672e05ad86de72535bd8dcad4cedR158) for yqla, etc. In the [PR](https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/601) I made those additional logic unreachable, because `handleUpdatingClusterState` returns a non-nil status (`ComponentStatusReady()`), the [updateQTState](https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/606/files#diff-8de26e1e52d4bdfeb4d2cea625c03860851a33787654f5951ea7719df9896adbR412)() function is never called, which means `prepareRestart()` is never executed. It lead to a situation when ytsaurus cluster stuck in Updating status indefinitely.
Current PR is fixing that issue by introducing [HasCustomUpdateState](https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/606/files#diff-df79516caab9f32d7a036081e9d1477e22679a71aa5dbf99f7d2889f081dc5a2R87) method of the Component Interface and adding small `if` [here](https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/606/files#diff-568998f3afe3b99cf89ffd554a0038cc17f8f747615d1a06fec01a336d20f779R175).
It's better to fix it that way then rolling back the [PR](https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/601), because `handleUpdatingClusterState` will be used later in the rollingUpdate logic. And also it reduce code duplication in qt, qa, yqla and sch components.

### DIFF
--- a/pkg/components/bundle_controller.go
+++ b/pkg/components/bundle_controller.go
@@ -89,3 +89,7 @@ func (bc *BundleController) Sync(ctx context.Context) error {
 func (bc *BundleController) doServerSync(ctx context.Context) error {
 	return bc.server.Sync(ctx)
 }
+
+func (bc *BundleController) HasCustomUpdateState() bool {
+	return false
+}
--- a/pkg/components/component.go
+++ b/pkg/components/component.go
@@ -84,6 +84,7 @@ type Component interface {
 	GetLabeller() *labeller.Labeller
 
 	GetCypressPatch() ypatch.PatchSet
+	HasCustomUpdateState() bool
 }
 
 // Following structs are used as a base for implementing YTsaurus components objects.
--- a/pkg/components/controller_agent.go
+++ b/pkg/components/controller_agent.go
@@ -91,3 +91,7 @@ func (ca *ControllerAgent) Sync(ctx context.Context) error {
 	_, err := ca.doSync(ctx, false)
 	return err
 }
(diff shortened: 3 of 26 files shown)


koct9i:
Let's look at overall update sequence design.
I don't like idea of tailoring code around pieces like handleUpdatingClusterState which isn't perfect as itself.

Right now update flow arranges sequence of update states which controls actions by independent components.

Activation logic is within update flow not inside components.
Update flow already must know details about components needs.

PR #610, opened 2025-12-05
Refactor component status, fix query tracker update

koct9i:
- **Added Query tracker update tests test to confirm we have a bag**
- **Refactor handling of server upgrade**


koct9i:
It is a complete mess

PR #634, opened 2025-12-18
Add tablet balancer to operator

ifsmirnov:


pkg/ytconfig/tablet_balancer.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"k8s.io/utils/ptr"
+
+	ytv1 "github.com/ytsaurus/ytsaurus-k8s-operator/api/v1"
+	"github.com/ytsaurus/ytsaurus-k8s-operator/pkg/consts"
+)
+
+type TabletBalancer struct {
+	WorkerThreadPoolSize int `yson:"worker_thread_pool_size"`
```
koct9i: I think "period" is worth to add here too And add both into instance spec.
For tiny test or demo clusters it is nice to have basic tuning for size and reaction time parameters.

PR #640, opened 2025-12-25
Added ca,ds and msc to the bulkUpdate mode

qurname2:
Adding ca, ds and msc into the bulkMode. Should be rebased from main after master's [PR](https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/630) is merged.

### DIFF
--- a/test/e2e/ytsaurus_controller_test.go
+++ b/test/e2e/ytsaurus_controller_test.go
@@ -2025,28 +2025,28 @@ exec "$@"`
 					sts := appsv1.StatefulSet{ObjectMeta: metav1.ObjectMeta{Namespace: namespace, Name: stsName}}
 					EventuallyObject(ctx, &sts, reactionTimeout).Should(HaveField("Spec.UpdateStrategy.Type", appsv1.OnDeleteStatefulSetStrategyType))
 
-					By("Verify StatefulSet has a new revision but pods stay on the old one")
-					EventuallyObject(ctx, &sts, reactionTimeout).Should(WithTransform(
-						func(current *appsv1.StatefulSet) bool {
-							return current.Status.CurrentRevision != "" &&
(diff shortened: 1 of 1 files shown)


.github/workflows/subflow_run_e2e_tests.yaml
```diff
@@ -23,13 +23,15 @@ jobs:
...
+      max-parallel: 6
       matrix:
         e2e:
           - name: "update basic"
             filter: "update && basic"
           - name: "update plan"
-            filter: "update && plan"
+            filter: "update && plan && !strategy"
+          - name: "update strategy"
+            filter: "update && plan && strategy"
```
koct9i: Oh. This escalates quickly.

Probably we could reuse one cluster and run several "updates" on it in random order.

I don't want "mock" clusters or invent fictional unit-test but bootstrapping took most of the time. Probably we could snapshot state and recreate it much faster.

test/e2e/ytsaurus_controller_test.go
```diff
@@ -1912,7 +1912,21 @@ exec "$@"`
...
 						ytsaurus.Spec.PrimaryMasters.InstanceCount = 3
+					case consts.MasterCacheType:
+						ytsaurus.Spec.MasterCaches = &ytv1.MasterCachesSpec{
+							InstanceSpec: ytv1.InstanceSpec{
+								Image:         ptr.To(testutil.YtsaurusImagePrevious),
+								InstanceCount: 3,
+							},
+						}
+					case consts.DiscoveryType:
+						ytsaurus.Spec.Discovery.Image = ptr.To(testutil.YtsaurusImagePrevious)
```
koct9i: Let's use some another reason for instance restart rather than image change.

We could treak "zero-diff" in config overrides or in instance environment.

This way we could reuse same cluster between tests.

PR #698, opened 2026-02-10
Implemented imageHeater as step 0 during cluster creation

qurname2:
Added `ClusterFeatures.EnableImageHeater` as a condition to run ImageHeater before any other components are being bootstrapped in order to pull required images on the nodes.


koct9i (approved):
LGTM for now

PR #708, opened 2026-02-17
Implemented rollingUpdate logic for hp/rp-proxies

qurname2:
[handleRollingUpdatingClusterState](https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/708/changes#diff-568998f3afe3b99cf89ffd554a0038cc17f8f747615d1a06fec01a336d20f779R352) implements rollingUpdate using maxUnavailable and compute the update budget dynamically on every reconcile cycle, advancing the StatefulSet partition by one pod per reconcile.
                                                                                                                                                                                
Algorithm:

  1. Init — if the STS still has the old spec (detected via [needUpdate](https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/708/changes#diff-568998f3afe3b99cf89ffd554a0038cc17f8f747615d1a06fec01a336d20f779R372)), apply the new spec with `partition = totalCount - 1` and `maxUnavailable` so that only the last pod is initially exposed to the update. Return and wait for the next reconcile. maxUnavailable is calculated as `int(server.getReplicaCount()) - server.getMinReadyInstanceCount()`

  2. Budget calculation — each reconcile reads the current STS status and computes:
  [inProgress](https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/708/changes#diff-568998f3afe3b99cf89ffd554a0038cc17f8f747615d1a06fec01a336d20f779R399)  = `max(0, (totalCount - partition) - updatedReplicas)`
  [effective](https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/708/changes#diff-568998f3afe3b99cf89ffd554a0038cc17f8f747615d1a06fec01a336d20f779R402)   = `max(totalCount - availableReplicas, inProgress)`
  [budget](https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/708/changes#diff-568998f3afe3b99cf89ffd554a0038cc17f8f747615d1a06fec01a336d20f779R403)      = `maxUnavailable - effective`
    - inProgress accounts for the timing window: pods that have been exposed by a prior partition lowering but haven't restarted yet.
    - effective takes the larger of actual unavailability and in-progress count to avoid double-counting.
  3. Advance — if budget > 0 and partition > 0, lower the partition by 1 and sync.
  4. Completion — when partition == 0 && updatedReplicas == totalCount, set the PodsUpdated condition.

  The Kubernetes StatefulSet controller's own maxUnavailable setting acts as a hard safety net: the operator drives the partition, but k8s will not restart more than maxUnavailable pods at once regardless.


pkg/components/helpers.go
```diff
@@ -335,6 +349,228 @@ func handleOnDeleteUpdatingClusterState(
...
+func setPodsUpdatedCondition(ctx context.Context, ytsaurus *apiproxy.Ytsaurus, cmp Component) {
+	ytsaurus.SetUpdateStatusCondition(ctx, metav1.Condition{
+		Type:    cmp.GetLabeller().GetPodsUpdatedCondition(),
+		Status:  metav1.ConditionTrue,
+		Reason:  consts.ConditionPodsUpdated,
+		Message: "All rolling batches were updated",
+	})
+}
+
+func getRollingBatchSize(ytsaurus *apiproxy.Ytsaurus, componentType consts.ComponentType, componentName string) (int32, bool) {
```
koct9i: I don't like such hanging single-use helpers.
They should be methods of object which manages this logic.

I suppose this must be component manager because it intakes update plan and arranges actual actions over selected set of components.

koct9i (changes requested):

koct9i (approved):
LGTM

I'm not sure how much "exhaustive" signal will be useful but ok for now.

PR #710, opened 2026-02-19
Cleanup logic of which component need update

koct9i:
- **Return component status from config map builder**
- **Check directly which components need update**
- **Update only components which needs update**
- **Remove lazy evaluation of component status**


qurname2 (approved):
lgtm!!!

PR #714, opened 2026-02-26
Fix Ytsaurus update stuck waiting for the `WaitingForTabletCellsRemoved` condition

aapurii:
## Description

During an upgrade of the operator chart from version 0.27.0 to 0.30.0, we encountered an issue where the CRD Ytsaurus got stuck in the `WaitingForTabletCellsRemoved` state.

## Problem analysis

The root cause was identified as the Bundle Controller (BC) being enabled for the `default` bundle. During the update process, the operator deletes tablet cells in two steps:
1. It removes cells via a simple `list + remove` operation on `//sys/tablet_cells`.
2. If the deletion requests succeed (i.e., no errors are returned), the operator repeatedly lists `//sys/tablet_cells` until the directory is empty.

However, cells managed by the Bundle Controller are automatically restored immediately after manual deletion. This causes the second step to hang indefinitely.

## Solution

To resolve this issue, we propose to temporarily disable Bundle Controller during tablet nodes update process. 

In details:

1. Before tablet cells removal:

- Save the current value of `//sys/@disable_bundle_controller` (it can be true, false, or unset).
- Disable Bundle Controller via `yt set //sys/@disable_bundle_controller %true` to prevent cells from recreating.

2. After tablet cells restored:

-  Restores original state of  the `//sys/@disable_bundle_controller`.


koct9i:
> 
> I think we should preserve the state of the bundle controller and restore it later because it could be unset (thus using the default value) or disabled by the user before the beginning of the update. So if we always enable the bundle controller after the update, it can lead to unexpected behaviour for the user.

No. This is bad engineering.  If bundle balalancer is in spec then it must be working. If you want disable it - then remove from spec, or add option to disable it's operations (I don't know reason why anybody might want that). Target state is defined by operator. If you don't want it - do not use operator.

> it can lead to unexpected behaviour for the user.

Is far as I know there is no documentation or example for that stuff - so there is no expected or unxpected behaviour

PR #729, opened 2026-03-10
Prepapred execNodes for rollingUpdate

qurname2:


pkg/components/exec_node.go
```diff
@@ -103,3 +109,79 @@ func (n *ExecNode) Sync(ctx context.Context, dry bool) (ComponentStatus, error)
...
+func (n *ExecNode) setDisableSchedulerJobs(ctx context.Context, podNameToDrain string) ComponentStatus {
+	ytClient, status := getYtClient(n.ytsaurusClient)
+	if status != nil {
+		return *status
+	}
+	stsName := n.GetLabeller().GetServerStatefulSetName()
+	cypressPath := consts.ComponentCypressPath(consts.ExecNodeType)
+
+	var instances []string
+	if err := ytClient.ListNode(ctx, ypath.Path(cypressPath), &instances, nil); err != nil {
```
koct9i: No. Listing and guessing by prefix is not an option.

This code should deduce exact name from index.
Labeller should have method for FQDN.
Exec node name also contains port which is known too.

Also need to think later about host-network.

pkg/components/exec_node.go
```diff
@@ -103,3 +109,93 @@ func (n *ExecNode) Sync(ctx context.Context, dry bool) (ComponentStatus, error)
...
+	for i := range totalCount {
+		if err := n.setExecNodeDisableSchedulerJobs(ctx, ytClient, i, false); err != nil {
+			return
+		}
+	}
+}
+
+func (n *ExecNode) execNodeInstancePath(ordinal int) ypath.Path {
+	instanceAddress := n.GetLabeller().GetInstanceAddressPort(ordinal, consts.ExecNodeRPCPort)
+	return ypath.Path(fmt.Sprintf("%s/%s", consts.ComponentCypressPath(consts.ExecNodeType), instanceAddress))
```
koct9i: `ypath.Path(consts.ComponentCypressPath(consts.ExecNodeType)).Child(instanceAddress)`

koct9i (approved):
LGTM

PR #739, opened 2026-03-13
set replication factor <3 in master init job

futujaos:
For clusters with replication factor less than 3 (i.e. with 1 or 2 data nodes), we have to set `//@replication_factor` so system dyntables will be created with correct replication factor, otherwise tablet nodes will fail to flush their chunks to data nodes.

It could be just `//sys/@replication_factor`, but we set for `//` for users convenience, so they don't have to make additional setup steps since all writes for clusters with `size(dnd) < 3` will be failed anyway.

### DIFF
--- a/pkg/components/master.go
+++ b/pkg/components/master.go
@@ -381,8 +381,8 @@ func (m *Master) createInitScript() (string, error) {
 		initCommands = append(
 			initCommands,
 			RunIfNonexistent(
-				"//sys/@replication_factor",
-				fmt.Sprintf("/usr/bin/yt set //sys/@replication_factor %d", m.cfgen.GetMaxReplicationFactor()),
+				"//@replication_factor",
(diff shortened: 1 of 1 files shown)


koct9i:
> > You could try to set default_table_replication_factor inside master config via config overrides.
> 
> Doesn't [this](https://github.com/ytsaurus/ytsaurus-k8s-operator/blob/6b355fa9b7b7a7c4ee92ad03031f6098969a0116/pkg/ytconfig/master.go#L88) code already do this?
> 
> > It it works we could add this into ytconfig generator for "Ephemera Cluster".
> 
> And I've already tried this using this patch in `ytserver-master.yson`, it doesn't help, unfortunately, system dyntables still will be created with rf=3.

Let's find out why config does not work. More likely this is default hardcoded in cli/sdk which is used inside init scripts. So operator could inject cli config into init jobs.

PR #740, opened 2026-03-13
Secondary Masters Maintenance [4]

### LINKED ISSUE #738: Add support for secondary masters
https://ytsaurus.tech/docs/en/admin-guide/cell-addition

spec

```yaml
primaryMasters:
  cellTag: 1
  instanceCount: 3
  affinity:
    podAntiAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        - labelSelector:
            matchLabels:
              yt_component: <name>-yt-master  # or ytsaurus.tech/cluster-name, app.kubernetes.io/component
          topologyKey: kubernetes.io/hostname # or topology.kubernetes.io/rack or topology.kubernetes.io/zone

secondaryMasters:
- cellTag: 2
  cellCount: 1 # spawn cellTag+0..n-1, default 1
  instanceCount: 3
  affinity:
    podAntiAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        - labelSelector:
            matchLabels:
              yt_component: <cluster>-yt-master-<cell>  # or ytsaurus.tech/cluster-name, app.kubernetes.io/component
            # or matchLabelKeys
          topologyKey: kubernetes.io/hostname # or topology.kubernetes.io/rack or topology.kubernetes.io/zone
```

primary masters pods: `ms-<index>`
secondary master pods: `ms-<cell>-<index>`

labels
```yaml
app.kubernetes.io/component: yt-master-<cell>
app.kubernetes.io/managed-by: ytsaurus-k8s-operator
app.kubernetes.io/name: yt-master
app.kubernetes.io/part-of: yt-<cluster>
yt_component: <cluster>-yt-master-<cell>
ytsaurus.tech/cluster-name: <clister>
ytsaurus.tech/cell-tag: <cell>
apps.kubernetes.io/pod-index: <index> # by sts
```

Nice to have: prefer collocate matching indexes between primary/secondary master:
https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#matchlabelkeys
by "apps.kubernetes.io/pod-index"

koct9i:
- **Validate that master cells are never removed and added only during maintenance**

```
    Allow adding only master cells with zero instance count,
    in this state is OK to remove them and change their roles.
    
    Changing instance count from zero to non-zero triggers actual extension.
    
    Also ensure that master cell does not lose quorum after changing instance count.
```

- **Introduce master maintenance workflow**

```
    Special multi-phase update flow for master cells maintenance.
    
    First of all during whole sequence cluster must stay in Maintenance:
    `clusterMaintenance: { shutdown: ExceptMasters }`.
    
    New secondary masters cells must be initially added with zero instance count.
    That saves their "unregistered" statues into resource status.
    Then switching from zero to non-zero instance count triggers actual extensions.
    
    Phases of master cells extension:
    - Reconfiguration: new master cells are created with 0 instances
    - Preparation: allow cells without roles, disable chunk refresh
    - Restart-1: old master cells build read-only snapshot and restart to accept new master cells
    - Acception: new master cells are allowed to register
    - Instantiation: new master instances are started for the first time
    - Registration: waiting for all new master cells register themselves
    - Restart-2: old and new masters build common read-only snapshot and restart
    - Settlement: verifing cell propagation and assigning roles for new master cells
    - Restart-3: all masters build read-only snapshot and restart to refresh cached cluster metadata
    - Completion: finish master cell extension, enable chunk refresh
    
    Undocumented pitfalls:
    - Old muster must knew unregistered master cells before their first start or all breaks
    - Snapshot building job must knew all registered cells and run after finishing registration
    - After registration masters needs to build common snapshot to leave "dynamic propagation"
    - Masters needs one more final restart after changing roles to refresh cached cluster metadata
    - Each read-only snapshot could stuck at switching leader and never recover themselves
```

Closes: #738


controllers/update_flow_steps.go
```diff
@@ -339,6 +346,59 @@ func buildFlowTree(componentManager *ComponentManager) *flowTree {
...
+	).chainIf(
+		updMaster,
+		st(ytv1.UpdateStateWaitingForMasterCellsExitReadOnly),
+	).chainIf(
+		// Waiting for registration of new master cells.
+		// Initial master cells settlement pass.
+		updMaster && componentManager.ytsaurus.IsUpdateStatusConditionTrue(consts.ConditionMasterCellsRegistration),
+		st(ytv1.UpdateStateWaitingForMasterCellsRegistration),
+	).chainIf(
+		// Checks that there is no dynamically propagated master cells then do master cells settlement:
```
achulkov2, line 408: In the guide there are some manual log checks suggested to be sure that new cells are ready to receive their roles (most importantly -- check that global objects have been replicated).
I didn't fully understand what we have replaced this with -- just the `//sys/@dynamically_propagated_masters_cell_tags == []` check?
Why is this enough?

Haven't dug into the corresponding master code yet, but I read the corresponding discussion in SIG Master chat and this choice doesn't seem to follow from it.

achulkov2 (dismissed):

controllers/update_flow_steps.go
achulkov2: Also from that discussion, aren't we supposed to call `reset_dynamically_propagated_master_cells` in some versions?

koct9i: "reset_dynamically_propagated_master_cells" is for future versions. We don't have it right now.

koct9i: > Why is this enough?

because measter call must be not "dynamically" propagated but "statically" or whatever they called. I.e. have common snapshot. That is enough.

manual log checks is about replication of common objects.
this part need to be reviewed. manual log checks sounds like bullshit. there should be proper way.

achulkov2: > manual log checks sounds like bullshit. there should be proper way.

Yes, that was my point, that right now we don't do any checks, and I don't think it is safe to assign new roles before that.

achulkov2: And same concern for "Leader active"/"Follower active" check.

koct9i: Assigning roles does nothing until cluster returns from maintenance.
Roles affects only master cell discovery when some actions needs cell with certain role.

kruftik (approved):

PR #747, opened 2026-03-24
API: add cluster maintenance

koct9i:
This group of option is designed for:
- switching cluster state into "Maintenance"
- shutdown components by scaling replicas count down to zero
- temporary switching off cluster features and automations

Do not remove and recover tablet cells when storage is shutdown
Tablets cannot be evicted when all data and tablet nodes are offline.
Perform tablet cell recover only if they were actually saved.

### DIFF
--- a/controllers/update_flow_steps.go
+++ b/controllers/update_flow_steps.go
@@ -263,7 +263,6 @@ func buildFlowTree(componentManager *ComponentManager) *flowTree {
 
 	updMaster := hasComponent(updatingComponents, consts.MasterType)
 	updTablet := hasComponent(updatingComponents, consts.TabletNodeType)
-	updMasterOrTablet := updMaster || updTablet
 	updDataNodes := hasComponent(updatingComponents, consts.DataNodeType)
 	updScheduler := hasComponent(updatingComponents, consts.SchedulerType)
 	updQueryTracker := hasComponent(updatingComponents, consts.QueryTrackerType)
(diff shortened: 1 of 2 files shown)


controllers/component_manager.go
```diff
@@ -39,6 +39,12 @@ type ComponentManagerStatus struct {
 	canUpdate    []ytv1.Component // Components with update allowed
 	cannotUpdate []ytv1.Component // Components with update blocked
 	nowUpdating  []ytv1.Component // Components updating right now
+
+	clusterMaintenance bool
+	mastersMaintenance bool
+	shutdownStorage    bool
```
qurname2, line 45: **storage** in my understanding it's data-nodes first of all, but the logic under it tells it's everything or everything except masters. Let's come up with better naming, please.

koct9i: Well. shutdownStorage is true  exactly when storage is shutdown.

When maintenance shutdown everything or when it shutdowns everything except master this includes data nodes. In other cases data nodes i.e. storage is not shutdown.

There is no mode shutdown just data nodes because it is useless - you cannot do anything useful in such state - everything besides masters and proxies are not operational. But there is mode to shutdown just tablet nodes. Because they are more complex and there are scenarios for fresh (re)start of all tablet nodes.

qurname2: Ok, sounds good to me in that case

PR #757, opened 2026-04-09
test/r8r: fix race in collecting events

koct9i:
There is no non-racy way to stop event broadcaster and watcher.


qurname2 (approved):
lgtm

PR #758, opened 2026-04-09
Refactor init jobs

koct9i:
- add script as text config generator to get correct configmap hash
- use one job for all operations by tracking reason
- remove in foreground previous job with reason not match
- restart job faster - after first back-off
- get completion condition right from update state name
- remove boilerplate code by adding method "RunScript"
- add proper logging and events for each completion


qurname2 (approved):
lgtm

PR #764, opened 2026-04-13
METAX GPU support

futujaos:


pkg/components/exec_node_base.go
```diff
@@ -191,6 +191,23 @@ func (n *baseExecNode) addCRIServiceSidecar(cri *ytconfig.CRIConfigGenerator, po
 		},
 	}
 
+	if cri.Runtime != nil {
```
koct9i, line 194: Move this into `cri.go` as method or helper

pkg/ytconfig/cri.go
```diff
@@ -249,16 +252,50 @@ func (cri *CRIConfigGenerator) getContainerdRuntimes() (runtimes map[string]any,
...
+	}
+	if cri.Runtime.Metax != nil {
+		gpuProvider = "metax"
+	}
+	if gpuProvider == "" {
+		return nil
+	}
+
+	return map[string]string{
+		// for gpuagent_runner.sh
```
koct9i: Then this should be added only for GPUInfoProvider == GPUInfoProviderGPUAgent

pkg/ytconfig/node.go
```diff
@@ -224,7 +224,9 @@ type GpuInfoSource struct {
 const GpuAgentPort = 23105
 
 type GpuManager struct {
-	GpuInfoSource GpuInfoSource `yson:"gpu_info_source"`
+	GpuInfoSource                        GpuInfoSource `yson:"gpu_info_source"`
+	GpuFlavor                            *string       `yson:"gpu_flavor,omitempty"`
+	UseGpuInfoProviderForDeviceDiscovery bool          `yson:"use_gpu_info_provider_for_device_discovery,omitempty"`
```
koct9i: I don't like this mess.
Could you explain what is this?
I though your bash script or gpu agent could provide all required information.

koct9i (changes requested):
see comments

pkg/ytconfig/cri.go
koct9i: How gpu agant is started?
IIRR we decices to leave this to entrypoint wrapper?
So you could just add "env YT_GPU_PROVIDEDR=whatever" into  entrypoint wrapper option.

pkg/ytconfig/node.go
futujaos: it's not for `gpuagent_runner.sh`, it's for exec node, to disable old behaviour with hard-coded `/dev/nvidia` checks and other NVIDIA-specific stuff, which is irrelevant for METAX and GpuAgent/k8s setup. We had to made hacks which emulate `/dev/nvidia` and other things, for METAX such hacks is not available at all. With this new exec nodes config old behaviour is disabled.

pkg/ytconfig/cri.go
futujaos: GpuAgent has `-gpu-provider` [argument](https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/gpuagent/cmd/agent/main.go#L40), it defines how it collects info about GPU

pkg/ytconfig/node.go
futujaos: here is this new [config fields](https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/server/lib/exec_node/config.cpp#L597)

koct9i: I still don't understand what is happening.
Could you please write proper pr description.

pkg/ytconfig/cri.go
futujaos: > Then this should be added only for GPUInfoProvider == GPUInfoProviderGPUAgent

fixed in [new PR](https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/793)

PR #765, opened 2026-04-13
EnableFullUpdate and UpdateSelector are removed

qurname2:


koct9i:
We could simply delete all these deprecated knobs all together.

koct9i:
add this into all sample configs:

```
  # Allow updating everything at once
  updatePlan:
    - class: Everything
```

koct9i (approved):

Issue #760, opened 2026-04-12
Skip EnableSafeMode when master are in onDelete/rollingUpdate update strategies


qurname2:
When masters are updating in the onDelete strategi, we need to skip the next update's steps:
- WaitingForSafeModeEnabled
- WaitingForSnapshots
- WaitingForMasterExitReadOnly
- WaitingForSafeModeDisabled
 
as it only increase cluster downtime from the user perspective.
How it can be done:

- Add `masterSkipsSafeMode bool` into [ComponentManagerStatus](https://github.com/ytsaurus/ytsaurus-k8s-operator/blob/main/controllers/component_manager.go#L28)
- Add helper `masterUsesNonBulkUpdateStrategy(updatePlan []ytv1.ComponentUpdateSelector)`
- Adjust update_flow_steps in chainIf(updMaster)

achulkov2:
Why do we need to add `masterSkipsSafeMode bool` into component manager? Shouldn't there be a better way to do update flow conditional on update strategy? Don't we have an option to skip some checks during rolling update, which should behave in a similar way?

qurname2:
We do have skipping checks during the rollingUpdate, but safe mode enabling is hardcoing into update_flow [here](https://github.com/ytsaurus/ytsaurus-k8s-operator/blob/main/controllers/update_flow_steps.go#L274). Anyway, I will evaluate the final solution more precisely.

PR #769, opened 2026-04-14
Added logic to skip tablet cells removal if tnd is in onDelete strategy

### LINKED ISSUE #494: Allow updating tablet nodes without deleting tablet cells
If dynamic tables are utilized quite heavily in terms of write throughput, deleting all tablet cells during updates takes a while, since tablet cell deletion unmounts all tables, which, in turn, waits for all dynamic stores to be flushed. 
This update mechanic was developed as a precaution, it is not strictly necessary, as long as tablet nodes are updated all at once. We can make this logic configurable in the spec, somewhere near the `updatePlan` logic.

qurname2:
closes #494

### DIFF
--- a/controllers/component_manager.go
+++ b/controllers/component_manager.go
@@ -409,7 +409,6 @@ func (cm *ComponentManager) areComponentPodsRemoved(component components.Compone
 func (cm *ComponentManager) applyUpdatePlan(updatePlan []ytv1.ComponentUpdateSelector) {
 	cm.status.canUpdate = nil
 	cm.status.cannotUpdate = nil
-	cm.status.removeTabletCellsOnUpdate = true
 	count := make([]int32, len(updatePlan))
 	for _, component := range cm.status.needUpdate {
 		index := -1
@@ -426,15 +425,28 @@ func (cm *ComponentManager) applyUpdatePlan(updatePlan []ytv1.ComponentUpdateSel
 		if index >= 0 {
 			count[index] += 1
 			cm.status.canUpdate = append(cm.status.canUpdate, component)
-			if component.Type == consts.TabletNodeType &&
-				updatePlan[index].Strategy != nil &&
-				updatePlan[index].Strategy.Type() == ytv1.ComponentUpdateModeTypeOnDelete {
-				cm.status.removeTabletCellsOnUpdate = false
-			}
 		} else {
 			cm.status.cannotUpdate = append(cm.status.cannotUpdate, component)
 		}
 	}
+}
+
+func shouldRemoveTabletCellsOnUpdate(updatePlan []ytv1.ComponentUpdateSelector, updatingComponents []ytv1.Component) bool {
+	for _, component := range updatingComponents {
+		if component.Type != consts.TabletNodeType {
+			continue
+		}
+		for _, selector := range updatePlan {
+			if !canUpdateComponent(selector, component) {
+				continue
+			}
+			if selector.Strategy != nil && selector.Strategy.Type() == ytv1.ComponentUpdateModeTypeOnDelete {
+				return false
+			}
+			break
+		}
+	}
+	return true
 }
 
 func canUpdateComponent(selector ytv1.ComponentUpdateSelector, component ytv1.Component) bool {
--- a/controllers/component_manager.go
+++ b/controllers/component_manager.go
@@ -45,6 +45,8 @@ type ComponentManagerStatus struct {
 	shutdownStorage    bool
 	shutdownTablets    bool
 	shutdownCompute    bool
+
+	removeTabletCellsOnUpdate bool // Skip tablet cell save/remove/recover when tablet nodes use OnDelete strategy
 }
 
 //nolint:cyclop //shush
@@ -407,6 +409,7 @@ func (cm *ComponentManager) areComponentPodsRemoved(component components.Compone
 func (cm *ComponentManager) applyUpdatePlan(updatePlan []ytv1.ComponentUpdateSelector) {
 	cm.status.canUpdate = nil
 	cm.status.cannotUpdate = nil
+	cm.status.removeTabletCellsOnUpdate = true
 	count := make([]int32, len(updatePlan))
 	for _, component := range cm.status.needUpdate {
 		index := -1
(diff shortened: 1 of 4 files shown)


koct9i:
I like the idea. But we need to think deeper.

Tablet cell removal is safest option when everything could be dangerously incompatible with previous state.
Kind of complete shutdown and cold start.

From the beginning this shouldn't be default option and adding flag to "disable" it isn't a good solution.
Spec here must reflect logic, not just expose bare wires to user.

api/v1/ytsaurus_types.go
```diff
@@ -838,6 +838,8 @@ type ClusterFeatures struct {
 	EnableImageHeater bool `json:"enableImageHeater,omitempty"`
 	// Enable logging anchor profiling by default for loggers that don't explicitly disable it.
 	EnableAnchorProfilingByDefault bool `json:"enableAnchorProfilingByDefault,omitempty"`
+	// Skip removing and recovering tablet cells during cluster update.
+	SkipTabletCellsRemovalOnUpdate bool `json:"skipTabletCellsRemovalOnUpdate,omitempty"`
```
achulkov2: Why is this part of such a global config? Can we move it to be a parameter of tablet node update? 
Unless it adds the option to remove tablet cells during all types of updates (but why).

Also, it is easier to understand `RemoveTabletCellsOnUpdate = true` then the double negative skip removal = false.

achulkov2 (changes requested):

api/v1/ytsaurus_types.go
qurname2: We had a talk and agreed that it would be better to move that logic under [ComponentManagerStatus](https://github.com/ytsaurus/ytsaurus-k8s-operator/blob/main/controllers/component_manager.go#L28). Decision about skipping tablet cells removal will be made automatically if updateStrategy is **onDelete**.
The same approach I suggested to use with master's safe_mode in the issue https://github.com/ytsaurus/ytsaurus-k8s-operator/issues/760

controllers/component_manager.go
```diff
@@ -403,6 +404,20 @@ func (cm *ComponentManager) applyUpdatePlan(updatePlan []ytv1.ComponentUpdateSel
 	}
 }
 
+func (cm *ComponentManager) computeRemoveTabletCellOnUpdate() bool {
```
koct9i: This function looks broken.

controllers/component_manager.go
```diff
@@ -35,10 +35,11 @@ type ComponentManagerStatus struct {
...
 
-	needUpdate   []ytv1.Component // Components in state NeedUpdate
-	canUpdate    []ytv1.Component // Components with update allowed
-	cannotUpdate []ytv1.Component // Components with update blocked
-	nowUpdating  []ytv1.Component // Components updating right now
+	needUpdate                []ytv1.Component // Components in state NeedUpdate
+	canUpdate                 []ytv1.Component // Components with update allowed
+	cannotUpdate              []ytv1.Component // Components with update blocked
+	nowUpdating               []ytv1.Component // Components updating right now
+	removeTabletCellsOnUpdate bool             // Skip tablet cell save/remove/recover when tablet nodes use OnDelete strategy
```
koct9i: you can add empty line above to keep allignment for other fields

controllers/ytsaurus_controller.go
```diff
@@ -259,6 +256,7 @@ func (r *YtsaurusReconciler) Sync(ctx context.Context, resource *ytv1.Ytsaurus)
 
 	case ytv1.ClusterStateUpdating:
 		cm.status.nowUpdating = ytsaurus.GetUpdatingComponents()
+		cm.status.removeTabletCellsOnUpdate = cm.computeRemoveTabletCellOnUpdate()
```
koct9i: this could be done inside applyUpdatePlan

like

```
func (cm *ComponentManager) applyUpdatePlan(updatePlan []ytv1.ComponentUpdateSelector) {
	cm.status.canUpdate = nil
	cm.status.cannotUpdate = nil
	count := make([]int32, len(updatePlan))
	for _, component := range cm.status.needUpdate {
		index := slices.IndexFunc(updatePlan, func(selector ytv1.ComponentUpdateSelector) bool {
			return canUpdateComponent(selector, component)
		})
		if index < 0 {
			cm.status.cannotUpdate = append(cm.status.cannotUpdate, component)
			continue
		}
		selector := updatePlan[index]
		if selector.Concurrency != nil && *selector.Concurrency <= count[index] {
			cm.status.cannotUpdate = append(cm.status.cannotUpdate, component)
			continue
		}
		count[index] += 1
		cm.status.canUpdate = append(cm.status.canUpdate, component)
		if component.Type == consts.TabletNodeType && selector.GetUpdateStrategyType() == ytv1.ComponentUpdateModeTypeBulkUpdate {
			cm.status.removeTabletCellsOnUpdate = true
		}
	}
}
```

qurname2: Good idea, will adjust the code

koct9i (approved):

controllers/component_manager.go
```diff
@@ -403,6 +405,24 @@ func (cm *ComponentManager) applyUpdatePlan(updatePlan []ytv1.ComponentUpdateSel
...
+			if !canUpdateComponent(selector, component) {
+				continue
+			}
+			if selector.Strategy != nil && selector.Strategy.Type() == ytv1.ComponentUpdateModeTypeOnDelete {
+				return false
+			}
+			break
+		}
+	}
+	return true
```
koct9i, line 449: If would be better in opposite way - return false when all updating tablet nodes have strategy ondetele, otherwise true

qurname2: Are you talking about potential case when we have multiply tablet node in the updatePlan and some of them is being updates using onDelete and some in bulk/rolling?

qurname2: But I don't think it is a good idea to have different strategies for multiple tablet nodes in one updatePlan.
I would rather forbid that on webhook layer

koct9i: Doesn't matter. LGTM for now. We'll definitely revisit this code later.

controllers/ytsaurus_controller.go
```diff
@@ -266,6 +266,7 @@ func (r *YtsaurusReconciler) Sync(ctx context.Context, resource *ytv1.Ytsaurus)
 
 	case ytv1.ClusterStateUpdating:
 		cm.status.nowUpdating = ytsaurus.GetUpdatingComponents()
+		cm.status.removeTabletCellsOnUpdate = shouldRemoveTabletCellsOnUpdate(resource.GetUpdatePlan(), cm.status.nowUpdating)
```
achulkov2, line 269: So it only applies in `OnDelete` mode?
But it is also useful in bulk update mode, if you want the update to go faster.
Just not by default, obvioiusly.

achulkov2 (changes requested):

controllers/ytsaurus_controller.go
achulkov2: We agreed to postpone this to another PR.

achulkov2 (approved):

PR #774, opened 2026-04-15
Refactor master quorum / update possibility check [3]

koct9i:
- **Refactor master cell quorum check**
- **Wait for component readiness in preparing state**
- **Refactor update possibility/health check**

### DIFF
--- a/pkg/checks/master.go
+++ b/pkg/checks/master.go
@@ -1,0 +1,98 @@
+package checks
+
+import (
+	"context"
+	"fmt"
+
+	"go.ytsaurus.tech/yt/go/ypath"
+	"go.ytsaurus.tech/yt/go/yt"
+
+	"github.com/ytsaurus/ytsaurus-k8s-operator/pkg/consts"
+)
+
+// NYT::NHydra::EPeerState
+type MasterState string
+
+const (
+	MasterStateLeading   MasterState = "leading"
+	MasterStateFollowing MasterState = "following"
+)
+
+// See: https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/server/lib/hydra/distributed_hydra_manager.cpp
+type MasterHydra struct {
+	State                MasterState `yson:"state"`
+	SelfID               int32       `yson:"self_id"`
+	LeaderID             int32       `yson:"leader_id"`
+	Voting               bool        `yson:"voting"`
+	Active               bool        `yson:"active"`
+	ActiveLeader         bool        `yson:"active_leader"`
+	ActiveFollower       bool        `yson:"active_follower"`
+	BuildingSnapshot     bool        `yson:"building_snapshot"`
+	EnteringReadOnlyMode bool        `yson:"entering_read_only_mode"`
+	LastSnapshotReadOnly bool        `yson:"last_snapshot_read_only"`
+	ReadOnly             bool        `yson:"read_only"`
+}
+
+type MasterAddressWithAttributes struct {
+	Address     string `yson:",value"`
+	Maintenance bool   `yson:"maintenance,attr"`
+}
+
+func MasterQuorumHealth(
+	ctx context.Context,
+	ytClient yt.Client,
+	mastersPath ypath.Path,
+	totalCount int32,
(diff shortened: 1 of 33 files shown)


pkg/components/ytsaurus_client.go
```diff
@@ -429,6 +359,7 @@ func (yc *YtsaurusClient) handleUpdatingState(ctx context.Context, dry bool) (Co
 
 func (yc *YtsaurusClient) Sync(ctx context.Context, dry bool) (ComponentStatus, error) {
 	var err error
+	logger := log.FromContext(ctx)
 	if hpStatus := yc.httpProxy.GetStatus(); !hpStatus.IsRunning() {
```
achulkov2, line 363: YtsaurusClient requires HTTP proxies. Cluster health checks require YtsaurusClient. So we cannot run them without alive HTTP proxies?
Might have always been a problem, but now this becomes more important, because master stuff happens with HTTP proxies down.

And it was a bad dependency always in any case.

pkg/components/ytsaurus_client.go
```diff
@@ -524,52 +465,73 @@ func (yc *YtsaurusClient) GetYtClient() yt.Client {
...
-	if msg != "" {
-		return false, msg, nil
+func (yc *YtsaurusClient) UpdateClusterHealthChecks(ctx context.Context) (err error) {
+	var updateIsPossible metav1.Condition
+	if !yc.shouldSkipCypressOperations() {
+		err = errors.Join(
+			yc.UpdateMasterQuorumCheck(ctx),
+			yc.UpdateCounterHealthCheck(ctx, consts.LostVitalChunksCountPath, consts.ConditionLostVitalChunksCheck),
+			yc.UpdateCounterHealthCheck(ctx, consts.QuorumMissingChunksCountPath, consts.ConditionQuorumMissingChunksCheck),
+			yc.UpdateTabletCellBundlesHealthCheck(ctx),
```
achulkov2, line 511: I am not sure this check will pass if we are extending masters and haven't deleted tablet cells before shutting down tablet nodes (which we don't have to do, and don't _want_ to do either). Maybe right now it works because we do delete tablet cells when shutting down tablet nodes => tablet cell bundles with zero cells are always healthy.

pkg/components/ytsaurus_client.go
```diff
@@ -524,52 +465,73 @@ func (yc *YtsaurusClient) GetYtClient() yt.Client {
...
+		)
```
achulkov2, line 512: And I generally dislike the fact that these checks are not skippable, I can easily imagine scenarios in which we want to be able to proceed with the update even when, say, LVC > 0.

pkg/checks/master.go
```diff
@@ -0,0 +1,98 @@
...
+			leaders = append(leaders, name)
+		case MasterStateFollowing:
+			followers = append(followers, name)
+		}
+	}
+
+	var note string
+	switch {
+	case len(maintenance) != 0:
+		note = "There is a master in maintenance"
```
achulkov2, line 83: So a master in maintenance mode makes any update impossible?
As we now run health checks periodically.

It is fine to update, idk, proxies, when a single master is in maintenance.
Maybe we can live without it in regular mode, but it can be very counter-productive if we are fixing an incident.

achulkov2 (changes requested):

pkg/components/ytsaurus_client.go
koct9i: Sure. You have discussed all these pre-checks a lot.
I don't see how this end this state. Let's discuss again how this should work.

pkg/checks/master.go
koct9i: Again. This was before my changes. I'm pretty sure you've reviewed this before.

pkg/components/ytsaurus_client.go
koct9i: Yes. This is stupidly bad design. Master simply must provide basic HTTP API.
But this should be changes in other repo.

achulkov2 (approved):

PR #781, opened 2026-04-17
Added runsMasterSafetySteps componentManager status in order to skip safety steps when we're in onDelete mode

### LINKED ISSUE #760: Skip EnableSafeMode when master are in onDelete/rollingUpdate update strategies

qurname2:
closes #760

### DIFF
--- a/controllers/component_manager.go
+++ b/controllers/component_manager.go
@@ -47,6 +47,7 @@ type ComponentManagerStatus struct {
 	shutdownCompute    bool
 
 	removeTabletCellsOnUpdate bool // Skip tablet cell save/remove/recover when tablet nodes use OnDelete strategy
+	runsMasterSafetySteps     bool // Run master pre/post-update safety steps (safe mode, snapshots, exit read-only) when master does NOT use OnDelete strategy
 }
 
 //nolint:cyclop //shush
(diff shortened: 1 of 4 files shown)


koct9i:
The name is confusing. It is always confusing when one piece of the code tweaks what other corner does instead of telling what is happening.

Let's call it "masterHotUpdate".

For update orchestration is does not matter how exactly update is implemented.
Important is that masters are supposed to continue operations without switching to read-only.

And inversion is good because allows to use "false" as  default safe state. This is important piece of go code design.

achulkov2:
I don't like either name. In the case of masters building snapshots is one of the characteristics of the update mode. E.g. you either "update via snapshots", which is the current bulk update flow, or you update via "rolling", which is the future flow. And OnDelete is a precursor to rolling. 
So there is no real need to make it configurable globally, this one can be derived form the update mode.

qurname2:
> So there is no real need to make it configurable globally, this one can be derived form the update mode.
There is no global configuration for the onDelete mode. The decision about setting master in read-only is getting derived from the updatePlan.

achulkov2:
Sorry, but "masterHotUpdate" also says absolutely nothing about what is actually going on.

The canonical name for updating masters by building snapshots and going into read-only is "via snapshot".
It is what has been used for years internally, also in support chats probably.

Let's call it masterUpdataViaSnapshots or something along those lines.

achulkov2:
Sorry, but "masterHotUpdate" also says absolutely nothing about what is actually going on.

The canonical name for updating masters by building snapshots and going into read-only is "via snapshot".
It is what has been used for years internally, also in support chats probably.

Let's call it masterUpdataViaSnapshots or something along those lines.

qurname2:
> Let's call it masterUpdataViaSnapshots or something along those lines.
I think that masterHotUpdate is more semantically correct here. 
- Operator doesn't care if there are snapshots or something else under the "cold" update mode. 
- That flag controls four steps and building a snapshot is only one of them.
- masterHotUpdate reads naturally as "not a hot update → run everything." and the default false value (cold update -> run all safety steps) of that bool variable looks safe enough. But with masterUpdataViaSnapshots it would be vice versa.

koct9i (approved):
LGTM

PR #790, opened 2026-04-23
Upgrade masters before starting other components

koct9i:
Right now we cannot start and upgrade components query-tracker which needs
data, tablet and exec nodes to migrate internal state. Current logic wait
when all components are started before leaving read-only state.

Only masters are required to switch masters from read-only state.
There is no reason to start any other components while masters are read-only:
they anyway could not bootstrap, register and become ready.

Create master pods at phase WaitingForMasterReady.

Other pods are created at phase WaitingForPodsCreation,
after WaitingForMasterExitReadOnly.


achulkov2 (approved):
LGTM.

PR #793, opened 2026-05-05
METAX GPU support

futujaos:
_I've messed up with original [PR](https://github.com/ytsaurus/ytsaurus-k8s-operator/pull/764) during rebase, so it is replacement for it._

* Added `Metax` section along with `Nvidia` in CRI spec
* `use_gpu_info_provider_for_device_discovery=true` now set in exec node static config for GpuAgent setup, it disables old behaviour with `/dev/nvidia*` discovery before even attempting to request devices info from GpuAgent; `gpu_flavor=other` is set for non-nvidia GPUs, it disables NVIDIA-specific code like reading driver version and other stuff; see more about this [config fields](https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/server/lib/exec_node/config.cpp#L597)
* `YT_GPU_PROVIDER` env variable is now set for GpuAgent runner script, since GpuAgent [have to know](https://github.com/ytsaurus/ytsaurus/blob/main/yt/yt/gpuagent/cmd/agent/main.go#L40) which GPUs it discovers to run vendor-specific discovery code


api/v1/ytsaurus_types.go
```diff
@@ -633,10 +633,15 @@ type GPUManagerSpec struct {
 
 type NvidiaRuntimeSpec struct{}
 
+type MetaxRuntimeSpec struct{}
+
 type JobRuntimeSpec struct {
 	// Use NVIDIA Container Runtime.
 	//+optional
 	Nvidia *NvidiaRuntimeSpec `json:"nvidia,omitempty"`
+	// Use NVIDIA Container Runtime.
```
koct9i: s/NVIDIA/MetaX/

pkg/ytconfig/cri.go
```diff
@@ -249,16 +252,53 @@ func (cri *CRIConfigGenerator) getContainerdRuntimes() (runtimes map[string]any,
...
 
-	if cri.Runtime != nil && cri.Runtime.Nvidia != nil {
-		runtimes[runtimeNameNvidia] = map[string]any{
-			"runtime_type": "io.containerd.runc.v2",
-			"sandbox_mode": "podsandbox",
-			"options": map[string]any{
-				"BinaryName": runtimePathNvidia,
-			},
+	if cri.Runtime != nil {
+		if cri.Runtime.Nvidia != nil {
```
koct9i, line 256: in Go this could be written as:

```
switch {
case cri.Runime == nil:
case cri.Runtime.Nvidia != nil:
  ...
case cri.Runtime.Metax != nil:
  ...
}
```

https://go.dev/ref/spec#Switch_statements

pkg/ytconfig/node.go
```diff
@@ -708,8 +710,12 @@ func getExecNodeServerCarcass(spec *ytv1.ExecNodesSpec, commonSpec *ytv1.CommonS
 	switch {
 	case spec.GPUManager != nil && spec.GPUManager.GPUInfoProvider != nil:
 		gpuInfoSource.Type = *spec.GPUManager.GPUInfoProvider
-	case spec.JobEnvironment != nil && spec.JobEnvironment.Runtime != nil && spec.JobEnvironment.Runtime.Nvidia != nil:
+	case spec.JobEnvironment != nil && spec.JobEnvironment.Runtime != nil && (spec.JobEnvironment.Runtime.Nvidia != nil || spec.JobEnvironment.Runtime.Metax != nil):
 		gpuInfoSource.Type = ytv1.GPUInfoProviderGPUAgent
+		c.ExecNode.GpuManager.UseGpuInfoProviderForDeviceDiscovery = true
+		if spec.JobEnvironment.Runtime.Nvidia == nil {
+			c.ExecNode.GpuManager.GpuFlavor = ptr.To("other")
+		}
```
koct9i: this is unreadable
add separate case for Metax

pkg/ytconfig/cri.go
```diff
@@ -249,16 +252,53 @@ func (cri *CRIConfigGenerator) getContainerdRuntimes() (runtimes map[string]any,
...
 	}
 
 	return runtimes, defaultRuntimeName
 }
+
+func (cri *CRIConfigGenerator) GetAdditionalEnvironmentVariables(spec *ytv1.ExecNodesSpec) map[string]string {
+	if cri.Runtime == nil {
+		return nil
+	}
+	if spec.GPUManager != nil && spec.GPUManager.GPUInfoProvider != ptr.To(ytv1.GPUInfoProviderGPUAgent) {
```
koct9i: Yep. This is wrong.

futujaos: why so? we always set gpuProvider to gpuAgent in `node.go` if it's not explicitly set in ytsaurus spec, so here i check if it's set and set not to gpuAgent

koct9i: look at copilot review for this line

you compare pointers, not values

also please click "resolve conversation" for fixed issues

koct9i (changes requested):

futujaos:
- Fixed issue with pointers (not value) comparison
- Added unit test for containerd config for METAX
- Added check for `GpuInfoProvider != GpuAgent` in `node.go` so new static config fields will apply even if user explicitly set provider to GpuAgent (remark from Copilot)

pkg/ytconfig/node.go
```diff
@@ -704,19 +706,24 @@ func getExecNodeServerCarcass(spec *ytv1.ExecNodesSpec, commonSpec *ytv1.CommonS
 		return c, err
 	}
 
-	gpuInfoSource := &c.ExecNode.GpuManager.GpuInfoSource
+	gpuManager := &c.ExecNode.GpuManager
 	switch {
```
koct9i, line 710: This is messy. And probably wrong. I'll look into this later.

koct9i (approved):

PR #796, opened 2026-05-08
 Fix HydraPersistenceUploader and Timbertruck mount paths

ilyaibraev:


api/v1/helpers.go
```diff
@@ -1,5 +1,18 @@
 package v1
 
+import (
+	corev1 "k8s.io/api/core/v1"
+)
+
+func FindFirstVolumeMount(volumeMounts []corev1.VolumeMount, name string) *corev1.VolumeMount {
```
koct9i: Nope. I'm not sure what is happening here but you definitely should add new type of location instead of that thing.

pkg/components/master.go
```diff
@@ -505,7 +506,12 @@ func (m *Master) runInitPhaseJobs(ctx context.Context, dry bool) (ComponentStatu
 	return m.initJob.RunScript(ctx, dry, "ClusterInitialization", m.scriptInitialization, nil)
 }
 
-func addHydraPersistenceUploaderToPodSpec(hydraImage string, podSpec *corev1.PodSpec, proxy string, secretKey string) {
+func addHydraPersistenceUploaderToPodSpec(hydraImage string, podSpec *corev1.PodSpec, proxy string, secretKey string, masterVolumeMounts []corev1.VolumeMount) {
+	masterDataMountPath := "/yt/master-data"
+	if vm := ytv1.FindFirstVolumeMount(masterVolumeMounts, "master-data"); vm != nil {
```
koct9i: =)

koct9i (changes requested):

api/v1/helpers.go
```diff
@@ -18,3 +25,44 @@ func FindAllLocations(locations []LocationSpec, locationType LocationType) []Loc
...
+		if strings.HasPrefix(normalizedPath, normalizedMount) {
+			if actual == nil || len(volumeMount.MountPath) > len(actual.MountPath) {
+				actual = volumeMount
+			}
+		}
+	}
+	return actual
+}
+
+func ResolveMountsForLocations(
```
koct9i: Let's move it out of api into pkg/components
and make more granular
if I understand this correctly you and mount into sidecar only some locations

then you could resolve location int volume+path add mount only volume subpath read-only/read-write.

so, then you don't need that de-duplication logic and this will not break if you need read-only and read-write locations which belongs to one volume - you just mount pieces of it twice.

api/v1/helpers.go
```diff
@@ -18,3 +24,19 @@ func FindAllLocations(locations []LocationSpec, locationType LocationType) []Loc
 	}
 	return result
 }
+
+func FindVolumeMountForPath(volumeMounts []corev1.VolumeMount, path string) *corev1.VolumeMount {
```
koct9i: I've asked to move this out of `api`. This function looks hairy. And probably there is something like that in k8s utils.

test/r8r/canondata/Components reconciler With all components Test/StatefulSet ms.yaml
```diff
@@ -165,12 +165,14 @@ spec:
         - mountPath: /config
           name: config-template
           readOnly: true
-        - mountPath: /yt/master-data
+        - mountPath: /yt/master-data/master-snapshots
           name: master-data
           readOnly: true
-        - mountPath: /yt/master-logs
-          name: master-logs
```
koct9i, line 168: uploader does not write own logs?

validators/ytsaurus_webhook.go
```diff
@@ -371,8 +381,29 @@ func (r *baseValidator) validateTimbertruckSpec(
...
+			"timbertruck",
+			parentPath,
+		)...)
+	}
+	return allErrors
+}
+
+func (r *baseValidator) validateRequiredLocations(
+	locations []ytv1.LocationSpec,
+	requiredLocations []ytv1.LocationType,
```
koct9i: make `requiredLocations` last va-args to make call more elegant.
and rename into requireLocations, and don't have to be method.

should be something like

allErrors = append(allErrors, requireLocations(
 parentPath.Child("locations"),
 locations,
 reson string,
 locationTypes...,
))

api/v1/helpers.go
koct9i: This code shouldn't do any normalization.

just is enough
`child == parent || strings.HasPrefix(child, parent+"/")`

iterate volumes backward - last mount always wins

And it is worth to add validation for for volume mount order and forbid mounts which completely covers previous mount

koct9i (changes requested):

koct9i:
and look here https://github.com/koct9i/ytsaurus-k8s-operator/pull/11

test/r8r/canondata/Components reconciler With all components Test/StatefulSet ms.yaml
ilyaibraev: Yes, [yt/docker/sidecars/hydra_persistence_uploader/bin/__main__.py](https://github.com/ytsaurus/ytsaurus/blob/main/yt/docker/sidecars/hydra_persistence_uploader/bin/__main__.py#L12-L17)

koct9i (approved):
LGTM

PR #810, opened 2026-06-01
Fix manual restart of init jobs

koct9i:
This must be allowed during initialization and all ready-to-update states,
because restart itself switches cluster into reconfiguration.

Do not remove init jobs automatically, because this triggers init job restart.


qurname2 (approved):
lgtm.

Issue #813, opened 2026-06-02
Remove insecure default admin password and token


koct9i:
YT spec could specify "adminCredentials" which defines admin login, password and token set during cluster initialization.

Without specifying this secret operator add superuser "admin", "password", "password".
*Which is stupidly insecure practice.*

Nowdays users could use token used for operator via http api or use "yt" cli via native api inside master container to create any users them want.

PR #814, opened 2026-06-02
Remove default insecure admin password and token

koct9i:
Do not create admin user when credentials are not specified.
Could be done any time using operator token or inside master pod.

Issue: #813


qurname2 (approved):
lgtm

PR #817, opened 2026-06-06
Support timbertruck for all components

savnadya:


pkg/components/server.go
```diff
@@ -197,17 +210,47 @@ func newServerConfigured(
 		generators...,
 	)
 
+	// Resolve timbertruck log delivery for this component. cfgen is required to compute the
```
koct9i: I don't like cfgen here. This seems layering violation.

main.go
```diff
@@ -72,6 +72,24 @@ func init() {
 	//+kubebuilder:scaffold:scheme
 }
 
+func setupWebhooks(mgr ctrl.Manager) error {
```
koct9i, line 75: This is just cleanup? Why it is here?

api/v1/ytsaurus_types.go
```diff
@@ -336,6 +336,10 @@ type TextLoggerSpec struct {
 type StructuredLoggerSpec struct {
 	BaseLoggerSpec `json:",inline"`
 	Category       string `json:"category,omitempty"`
+	// EnableDelivery turns on timbertruck delivery of this structured log to the cluster log storage.
+	// Requires a timbertruck image to be configured (spec.timbertruck.image or, for masters, primaryMasters.timbertruck.image).
+	//+optional
+	EnableDelivery *bool `json:"enableDelivery,omitempty"`
```
koct9i, line 342: If this is exclusively related to "timbertrack" then name it "enableTimbertruckDelively".

api/v1/ytsaurus_types.go
```diff
@@ -478,6 +482,9 @@ type HydraPersistenceUploaderSpec struct {
 	Image *string `json:"image,omitempty"`
 }
 
+// TimbertruckSpec configures the timbertruck sidecar that delivers structured logs to the
+// cluster log storage. It is used both cluster-wide (spec.timbertruck) and as a per-component
+// override (e.g. primaryMasters.timbertruck).
 type TimbertruckSpec struct {
 	Image         *string `json:"image,omitempty"`
 	DirectoryPath *string `json:"directoryPath,omitempty"`
```
koct9i, line 506: Please document fields here.

pkg/components/server.go
savnadya: I have removed the unnecessary YT_PROXY, it is no longer needed

koct9i (approved):
good enough

PR #819, opened 2026-06-15
Fix "Conflicting profiling tags" alert at tablet nodes

ifsmirnov:
If bundle controller is not present, set
"/cluster_node/deduce_profiling_tag_from_bundle_name" option to the node config.


koct9i (approved):
LGTM

PR #828, opened 2026-07-20
Add option to configure init container resources

Gufran:
This change allows configuring resources for init containers which is needed to place the containers in guaranteed cgroups.

Issue: #829


pkg/components/chyt.go
```diff
@@ -46,7 +46,7 @@ func NewChyt(cfgen *ytconfig.NodeGenerator, chyt *apiproxy.Chyt, ytsaurus *ytv1.
 			cfgen.GetNativeClientConfig,
 			&ytsaurus.Spec.CommonSpec,
 			&ytsaurus.Spec.PodSpec,
-			&ytv1.InstanceSpec{},
```
koct9i, line 49: No. Not "PrimaryMasters".

This is "InitJob". That's separate pod. It don't need "guaranteed" QoS

pkg/components/server.go
```diff
@@ -544,6 +544,7 @@ func (s *serverImpl) rebuildStatefulSet() *appsv1.StatefulSet {
 		Tolerations:  getTolerationsWithDefault(s.instanceSpec.Tolerations, s.commonPodSpec.Tolerations),
 		DNSConfig:    ptrDefault(s.instanceSpec.DNSConfig, s.commonPodSpec.DNSConfig),
 	}
+	setContainerResources(podSpec.InitContainers, podSpec.Containers[0].Resources)
```
koct9i, line 547: Probably it is the only place which needs change.
May be for exec nodes, but I'm not sure.

koct9i:
Validation part #835

koct9i (approved):

PR #830, opened 2026-07-21
Start caches/discovery/tablets after master [1]

koct9i:
- **Start master caches only when masters are running**
- **Start discovery service only when masters are running**
- **Start tablet nodes only when masters are running**

Start master caches only when masters are running
Otherwise master caches and nodes could see and cache non-final cluster metadata.


qurname2 (approved):
lgtm

PR #832, opened 2026-07-21
Preserve YSON field order and use go-difflib instead go-cmp for reporting config diff

### LINKED ISSUE #748: Presence of config overrides for a component leads to change in config structure and update of that component
1. Create ytsaurus cluster
2. Add in config overrides something like
  ```
    ytserver-query-tracker.yson: |
      {}
  ```
3. See how operator changed config and wants to update query tracker (or any other component)

I think we need some normalisation of resulting config, so this does not happen and resulting config does not change on some order/noop changes

koct9i:
- test/r8r: add ordered test-cases over same cluster
- Use go-difflib instead go-cmp for reporting config diff
- Preserve YSON field order when applying overrides

Closes: #748


koct9i:
This makes #748 obvious

PR #836, opened 2026-07-21
Disable stockpile by default

koct9i:
Stop flood in debug log.

It is not supported in mainline linux and off by default since 25.3


qurname2 (approved):
lgtm.

PR #843, opened 2026-07-28
Fix init job (re)start consistency

koct9i:
Set condition reason between removing previous and starting new job.
This prevents restating already started job if status is stale.


qurname2 (approved):
lgtm

PR #847, opened 2026-08-04
Support yson structured logs and multi-category filter in timbertruck

tinarsky:
Needed to deliver the scheduler event log: it is yson and spans two categories.

- Structured loggers with `format: yson` now go to timbertruck's `yson_logs` section. They were emitted as `json_logs`, whose validator dropped every line — delivery looked healthy while nothing arrived. Requires a timbertruck image that understands `yson_logs`.
- A structured logger can select several categories via `categoriesFilter`. It replaces `category` rather than extending it: a webhook requires exactly one of the two, and rejects a filter with no type or values, which would otherwise leave the rule matching every category. `category` keeps working and was already optional in the schema, so existing specs are untouched.

A `format: plain_text` structured logger still goes to `json_logs` — pre-existing, left as is.

### DIFF
--- a/api/v1/ytsaurus_types.go
+++ b/api/v1/ytsaurus_types.go
@@ -335,9 +335,11 @@ type TextLoggerSpec struct {
 
 type StructuredLoggerSpec struct {
 	BaseLoggerSpec `json:",inline"`
-	Category       string `json:"category,omitempty"`
-	// Additional include/exclude category filter applied on top of category.
-	// Type must be set.
+	// Single category to log. Exactly one of category and categoriesFilter must be set.
+	//+optional
+	Category string `json:"category,omitempty"`
+	// Category filter for logs spanning several categories, as for text loggers.
+	// Type must be set. Exactly one of category and categoriesFilter must be set.
 	//+optional
 	CategoriesFilter *CategoriesFilter `json:"categoriesFilter,omitempty"`
 	// EnableDelivery turns on timbertruck delivery of this structured log to the cluster log storage.
--- a/api/v1/ytsaurus_types.go
+++ b/api/v1/ytsaurus_types.go
@@ -337,8 +337,7 @@ type StructuredLoggerSpec struct {
 	BaseLoggerSpec `json:",inline"`
 	Category       string `json:"category,omitempty"`
 	// Additional include/exclude category filter applied on top of category.
-	// Lets one logger span several categories, e.g. the scheduler event log is written
-	// under both SchedulerStructuredLog and SchedulerEventLog. Type must be set.
+	// Type must be set.
 	//+optional
 	CategoriesFilter *CategoriesFilter `json:"categoriesFilter,omitempty"`
 	// EnableDelivery turns on timbertruck delivery of this structured log to the cluster log storage.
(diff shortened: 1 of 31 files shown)


pkg/ytconfig/timbertruck.go
```diff
@@ -10,8 +10,10 @@ import (
 )
 
 type TimbertruckConfig struct {
-	WorkDir  string                     `json:"work_dir" yson:"work_dir"`
-	JsonLogs []TimbertruckJsonLogConfig `json:"json_logs" yson:"json_logs"`
+	WorkDir string `json:"work_dir" yson:"work_dir"`
+	// Timbertruck validates each section in its own format and drops unparsable lines.
+	JsonLogs []TimbertruckJsonLogConfig `json:"json_logs,omitempty" yson:"json_logs,omitempty"`
+	YsonLogs []TimbertruckJsonLogConfig `json:"yson_logs,omitempty" yson:"yson_logs,omitempty"`
```
savnadya: TimbertruckYsonLogConfig?

api/v1/ytsaurus_types.go
```diff
@@ -336,6 +336,10 @@ type TextLoggerSpec struct {
 type StructuredLoggerSpec struct {
 	BaseLoggerSpec `json:",inline"`
 	Category       string `json:"category,omitempty"`
+	// Additional include/exclude category filter applied on top of category.
+	// Type must be set.
+	//+optional
+	CategoriesFilter *CategoriesFilter `json:"categoriesFilter,omitempty"`
```
savnadya, line 344: Now it seems strange that "Category" is required

pkg/components/pods_manager.go
```diff
@@ -60,6 +61,11 @@ func podContainerIsReady(pod *corev1.Pod, name string) bool {
 func podContainerNames(pod *corev1.Pod, names []string) []string {
 	if len(names) == 0 {
 		for _, ct := range pod.Spec.Containers {
+			// Log delivery must not gate readiness: kubelet restarts the sidecar on its own,
+			// while a failing one would otherwise stall the whole cluster update.
+			if ct.Name == consts.TimbertruckContainerName {
```
savnadya: no, this is a very bad hack

savnadya (dismissed):

pkg/ytconfig/timbertruck.go
tinarsky: Done: split into TimbertruckJsonLogConfig and TimbertruckYsonLogConfig over a shared TimbertruckBaseLogConfig, mirroring timbertruck's own JSONLogConfig/YSONLogConfig.

api/v1/ytsaurus_types.go
tinarsky: category is optional now, and categoriesFilter replaces it instead of extending it — a webhook requires exactly one of the two. Existing specs are untouched; category was already optional in the schema.

pkg/components/pods_manager.go
tinarsky: Dropped, pods_manager.go is back to main — the sidecar keeps gating readiness.

savnadya:
LGTM

api/v1/ytsaurus_types.go
```diff
@@ -335,7 +335,13 @@ type TextLoggerSpec struct {
 
 type StructuredLoggerSpec struct {
```
koct9i, line 336: Add above struct 

// +kubebuilder:validation:XValidation:rule="has(self.foo) != has(self.bar)",message="exactly one of foo or bar must be set"

and then drop code from webhook

koct9i (approved):

PR #857, opened 2026-08-18
Pin Telegram action, gate runs on secrets, and set workflow permissions

koct9i:
- Ensure workflows run safely only when required secrets are present and to avoid unintended permission elevation by explicitly setting `permissions` and gating notification/release steps on secret presence.
- Replace floating `appleboy/telegram-action@master` with a pinned, immutable reference to avoid unexpected behavior from upstream changes.
- Prevent leaking credentials or failing releases when Docker Hub secrets are absent by conditionally running Docker login and release steps.

- Added `permissions: {}` to `issue_notifications.yaml` and `pr_notifications.yaml` to tighten default workflow permissions and added explicit `permissions` for `subflow_release.yaml` with `contents: read` and `packages: write`.
- Replaced `appleboy/telegram-action@master` with a pinned commit `appleboy/telegram-action@221e6b684967abe813051ee4a37dd61770a83ad3` and added `if` checks plus `HAS_TELEGRAM_SECRETS` env expressions to skip Telegram steps when required secrets are missing in `issue_notifications.yaml`, `pr_notifications.yaml`, and `subflow_release.yaml`.
- Added conditional gating for Docker Hub operations in `subflow_release.yaml` by introducing `HAS_DOCKER_HUB_SECRETS` env checks, using `printenv DOCKER_HUB_PASSWORD` for login input, and skipping the `Release` step when Docker Hub credentials are not present.

### DIFF
--- a/.github/workflows/subflow_release.yaml
+++ b/.github/workflows/subflow_release.yaml
@@ -8,6 +8,10 @@ on:
         required: false
         type: string
 
+permissions:
+  contents: read
+  packages: write
+
 jobs:
   release:
     name: Run release
@@ -40,9 +44,14 @@ jobs:
           make build
 
       - name: Docker Hub login
+        if: env.HAS_DOCKER_HUB_SECRETS == 'true'
+        env:
+          HAS_DOCKER_HUB_SECRETS: ${{ secrets.DOCKER_HUB_LOGIN != '' && secrets.DOCKER_HUB_PASSWORD != '' }}
+          DOCKER_HUB_LOGIN: ${{ secrets.DOCKER_HUB_LOGIN }}
+          DOCKER_HUB_PASSWORD: ${{ secrets.DOCKER_HUB_PASSWORD }}
         shell: bash
         run: |
-          echo '${{ secrets.DOCKER_HUB_PASSWORD }}' | docker login --username ${{ secrets.DOCKER_HUB_LOGIN}} --password-stdin
+          printenv DOCKER_HUB_PASSWORD | docker login --username "${DOCKER_HUB_LOGIN}" --password-stdin
 
       # Uses the `docker/login-action` action to log in to the Container registry
       # using the account and password that will publish the packages.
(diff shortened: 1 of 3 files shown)


kruftik (approved):

.github/workflows/subflow_release.yaml
```diff
@@ -55,9 +64,10 @@ jobs:
           password: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Release
+        if: env.HAS_DOCKER_HUB_SECRETS == 'true'
         shell: bash
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          HAS_DOCKER_HUB_SECRETS: ${{ secrets.DOCKER_HUB_LOGIN != '' && secrets.DOCKER_HUB_PASSWORD != '' }}
```
savnadya, line 70: Why do you just remove a GITHUB_TOKEN?

savnadya (changes requested):

.github/workflows/subflow_release.yaml
koct9i: Because it is already in default env

https://docs.github.com/en/enterprise-cloud@latest/codespaces/developing-in-a-codespace/default-environment-variables-for-your-codespace#list-of-default-environment-variables

savnadya (approved):

.github/workflows/subflow_release.yaml
koct9i: Well, or maybe no. Ok. I'll revert this line.

https://docs.github.com/en/actions/reference/workflows-and-actions/variables#default-environment-variables

koct9i: Actually this step does not need GITHUB_TOKEN at all

koct9i: yeah, it is use by previous step - `docker/login-action`

PR #859, opened 2026-08-18
Сonfigurable logs for native clients

epsilond1:
This change will able to debug problems in native clients, for example, the custom-yt-init-job can failing quietly now (incident with [bug](https://github.com/ytsaurus/ytsaurus/commit/9b98ab7ed920efbab9048368eb53847ac48c00bd) in ares), and also other calls yt tool, these make connections via native protocols.


koct9i (changes requested):
CRD impact is too huge

Issue #820, opened 2026-06-17
Data node chunk location quota should be unlimited by default


koct9i:
This code should not use PVC storage REQEUST as default quota limit.
Data should use all available space IF quota for location is not defined explicitly.
It may use PVC storage LIMIT as default quota limit.

Reported-by: @Gufran

PR #860, opened 2026-08-18
Default location QUOTA should came from volume storage LIMIT

koct9i:
If both volume storage limit and location quota are not fined then
location should use all available space.

But default low watermark should come from volume storage request.

This is required for full utilization when exact disk sizes are unknown.

Issue: #820


qurname2 (approved):
lgtm.

Issue #788, opened 2026-04-22
Set memory_limits in configs for http and rpc proxies


Krock21:
http proxies don't have this limit
rpc proxies default it to 20gb in code
https://github.com/ytsaurus/ytsaurus/blob/de905ed99a09801421894cb9e00cca8edd0ca47b/yt/yt/server/http_proxy/config.cpp#L135
https://github.com/ytsaurus/ytsaurus/blob/de905ed99a09801421894cb9e00cca8edd0ca47b/yt/yt/server/rpc_proxy/config.cpp#L174

I think it makes sense to do the same logic as we do with node resource_limits.total_memory for cluster nodes and set it in configs for proxies

PR #861, opened 2026-08-18
Set memory limit in HTTP and RPC proxy configs

koct9i:
Proxies needs to know at least total size of available memory.

Issue: #788


qurname2 (approved):
lgtm.

PR #864, opened 2026-08-19
Requeue reconcile after update step progress and fix race at UpdateIsPossible

koct9i:
Reconciler should return RequeueAfter = DefaultClusterStatusPollPeriod only when it is stuck.

Fix race at checking UpdateIsPossible
Update step shouldn't pass until it see own successful result.


qurname2 (approved):
LGTM.

Issue #869, opened 2026-08-20
Make RemoveTabletCells concurrent


qurname2:
Right now, RemoveTabletCells is removing one-by-one what is taking pretty big time on a large YT cluster.
We can speed up the process by adding concurrency here.

koct9i:
Nope, tablet cell remove just initiates decommission. Remove itself isn't blocking operation.

PR #873, opened 2026-08-22
Disable tablet cell decommission while removing them

koct9i:
This prevents migration of tablets into remaining tablet cells.
Which is potentially leads to O(N^2) migrations.

Issue: #869


qurname2 (approved):
lgtm.

PR #875, opened 2026-08-24
Add tablet-node option to set tablet-cell slots

koct9i:
- **test/r8r: cover tablet nodes**
- **Use envtest for tablet node component tests**
- **API: add tablet-node option to set count of tablet-cell slots**
- **Make disabled caches in static config more explicit**


Krock21 (approved):
LGTM

